PR #73, opened 2025-10-09
feature(common): Build UpcomingQuizCard component

kellyalejandra861:
image

https://ditmar.github.io/leetcode-spa/feature-leetcode-12/?path=/story/components-upcomingquizcard--default

Add date-fns to parse/format ISO date strings in UpcomingQuizCard utils/hook (runtime use).
It is used at runtime to format and parse dates within the component UpcomingQuizCard.
It should remain as a regular dependency ( dependencies) , not as devDependency, since it runs in production.
The version is fixed ( 4.1.0), complying with the project rules on exact versions.

test on terminal
Captura de pantalla 2025-10-26 085320

### DIFF
--- a/src/components/UpcomingQuizCard/UpcomingQuizCard.styles.ts
+++ b/src/components/UpcomingQuizCard/UpcomingQuizCard.styles.ts
@@ -5,6 +5,18 @@ export const StyledCard = styled(Card)(({ theme }) => ({
   textAlign: 'center',
   width: theme.spacing(52.75),
   height: theme.spacing(38.375),
+  margin: '0 auto', // 🔥 siempre centrado
+
+  [theme.breakpoints.down('sm')]: {
+    width: '90%',
+    height: 'auto',
+    padding: theme.spacing(3),
+    margin: '20px auto', // 🔥 centrado en vista móvil
+    display: 'flex',
+    flexDirection: 'column',
+    alignItems: 'center', // 🔥 contenido centrado
+    justifyContent: 'center',
+  },
 }));
 
 export const DateContainer = styled(Box)(({ theme }) => ({
@@ -12,6 +24,11 @@ export const DateContainer = styled(Box)(({ theme }) => ({
   display: 'flex',
   flexDirection: 'column',
   alignItems: 'center',
+
+  [theme.breakpoints.down('sm')]: {
+    margin: theme.spacing(1.5, 0),
+    justifyContent: 'center',
+  },
 }));
 
 export const TypographyButton = styled(Typography)(({ theme }) => ({
@@ -19,34 +36,52 @@ export const TypographyButton = styled(Typography)(({ theme }) => ({
   color: theme.palette.common.white,
   textTransform: 'none',
   fontWeight: theme.typography.fontWeightMedium,
-  fontSize: theme.typography.pxToRem(17),
+  fontSize: theme.typography.pxToRem(16),
+  width: theme.spacing(11.875),
+  height: theme.spacing(2.25),
+  [theme.breakpoints.down('sm')]: {
+    fontSize: theme.typography.pxToRem(15),
+    textAlign: 'center',
+  },
 }));
 
 export const RegisterButton = styled(Button)(({ theme }) => ({
-  width: theme.spacing(31.125),
-  height: theme.spacing(6.75),
-
+  width: theme.spacing(35.125),
+  height: theme.spacing(8.25),
   borderRadius: theme.spacing(3.75),
   opacity: 1,
-
   gap: theme.spacing(1.25),
-  background: 'linear-gradient(60deg, #B33DEB 13.4%, #DE8FFF 86.6%)',
+  background: 'linear-gradient(60deg, #B23DEB 13.4%, #DE8FFF 86.6%)',
 
-  '&:hover': { opacity: 0.9, background: 'linear-gradient(60deg, #A030D9 13.4%, #D080F2 86.6%)' },
+  '&:hover': {
+    opacity: 0.9,
+    background: 'linear-gradient(60deg, #A030D9 13.4%, #D080F2 86.6%)',
+  },
+
+  [theme.breakpoints.down('sm')]: {
+    width: '100%',
+    height: theme.spacing(6),
+    borderRadius: theme.spacing(3),
+    display: 'flex',
+    justifyContent: 'center',
+  },
 }));
 
 export const StyleTitle = styled(Typography)(({ theme }) => ({
   fontFamily: `Syne, ${theme.typography.fontFamily}`,
   fontWeight: theme.typography.fontWeightMedium,
   fontSize: theme.typography.pxToRem(17),
-
   textAlign: 'center',
-  color: theme.palette.grey[900] + 'CC',
   opacity: 1,
-
   width: theme.spacing(28.75),
   height: theme.spacing(2.5),
-
   margin: `0 auto`,
   display: 'block',
+
+  [theme.breakpoints.down('sm')]: {
+    fontSize: theme.typography.pxToRem(15),
+    width: '100%',
+    height: 'auto',
+    textAlign: 'center',
+  },
 }));
--- a/src/components/UpcomingQuizCard/UpcomingQuizCard.tsx
+++ b/src/components/UpcomingQuizCard/UpcomingQuizCard.tsx
@@ -1,4 +1,4 @@
-import { Typography, CardContent, Avatar } from '@mui/material';
+import { Typography, Avatar } from '@mui/material';
 import React from 'react';
 
 import Calendar from './icon/Calendar.png';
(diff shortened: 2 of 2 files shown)


src/components/Button/Button.tsx
```diff
@@ -0,0 +1,56 @@
...
+      disabled={disabled || loading}
+      onClick={handleClick}
+      data-testid={dataTestId}
+      $variant={variant}
+      $size={size}
+      $shape={shape}
+      $loading={loading}
+      $fullWidth={fullWidth}
+    >
+      {loading && <LoadingSpinner size={size === 'small' ? 16 : size === 'large' ? 24 : 20} />}
```
Ditmar: LoadingSpinner size is computed via ternary expressions with magic numbers (16, 20, 24). Consider extracting these numbers into SIZE_CONFIGS so sizes are consistent and theme-aligned.

**Sugerencia:**
```typescript
Use const spinnerSize = size === 'small' ? SIZE_CONFIGS.small.spinnerSize : size === 'large' ? SIZE_CONFIGS.large.spinnerSize : SIZE_CONFIGS.medium.spinnerSize; and define spinnerSize in constants (or theme).
```

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,56 @@
...
+      <ButtonContent $loading={loading}>
+        {startIcon && <span>{startIcon}</span>}
```
Ditmar: startIcon and endIcon are wrapped inside <span>{startIcon}</span>. If the passed node already includes semantics or aria attributes, the extra wrapper could interfere with styling or accessibility. Also MUI Button supports startIcon/endIcon props by default; consider reusing those instead of custom markup.

**Sugerencia:**
```typescript
Either pass startIcon and endIcon to the underlying MUI Button (it supports startIcon/endIcon props), or ensure wrapper has aria-hidden if purely decorative: <span aria-hidden>{startIcon}</span>.
```

package.json
```diff
@@ -28,6 +28,7 @@
     "@mui/icons-material": "7.3.4",
     "@mui/material": "7.3.4",
     "astro": "5.13.9",
+    "date-fns": "4.1.0",
```
Ditmar: A new runtime dependency (date-fns@4.1.0) was added. Please include a short justification in the PR description explaining why date-fns is necessary (used at runtime for formatting/parsing dates). Confirm this should be a regular dependency (not devDependency). The version is pinned (no caret) which matches project rules — good. Also ensure the project licensing/policy allows this package and that tree-shaking/size impact was considered.

**Sugerencia:**
```typescript
Keep date-fns as a runtime dependency only if the component formats/parses dates at runtime (it does). Add a one-line justification in PR description like: "Add date-fns to parse/format iso date strings in UpcomingQuizCard utils/hook (runtime use)."
```

src/components/UpcomingQuizCard/UpcomingQuizCard.constants.ts
```diff
@@ -0,0 +1 @@
+export const DEFAULT_DATE_FORMAT = 'dd MMM, yyyy';
```
Ditmar: DEFAULT_DATE_FORMAT is hard-coded to 'dd MMM, yyyy'. Date presentation is locale-sensitive; using a fixed pattern may not be appropriate for all locales. Also date formats should be localizable or derived from the user's locale when required.

**Sugerencia:**
```typescript
Consider storing only a default pattern but document that callers can pass a locale. Alternatively, use date-fns localization APIs (format with a locale option) or expose an optional locale parameter in formatDate/useUpcomingQuizCard so UI can be localized.
```

src/components/UpcomingQuizCard/UpcomingQuizCard.hook.ts
```diff
@@ -0,0 +1,29 @@
+import { useMemo } from 'react';
+
+import { formatDate } from './UpcomingQuizCard.utils';
+
+export const useUpcomingQuizCard = (date: string | Date) => {
+  const formattedDate = useMemo(() => {
+    const normalizedDate = normalizeDate(date);
+    const nextDay = new Date(normalizedDate);
```
Ditmar: The hook constructs a Date from the normalized date then unconditionally adds one day (nextDay.setDate(nextDay.getDate() + 1)) before formatting. This produces an incorrect date (off-by-one). Adjusting dates by +1 to compensate for timezone/parsing issues is a fragile approach and likely a bug.

**Sugerencia:**
```typescript
Do not add +1 day. Use a robust parsing approach (parseISO from date-fns) and then format the parsed date directly. Example fix:

const normalizedDate = normalizeDate(date);
return formatDate(normalizedDate);

Better: normalizeDate should return a valid Date (using parseISO for ISO strings) and formatDate should accept that Date. If timezone normalization is required, handle it explicitly (e.g., startOfDay or use UTC helpers) and document why.
```

src/components/UpcomingQuizCard/UpcomingQuizCard.stories.tsx
```diff
@@ -0,0 +1,25 @@
...
+  component: UpcomingQuizCard,
+  args: {
+    title: 'Upcoming Quiz Competition',
+    date: '2025-08-12',
+    description: 'Join our next exciting quiz and test your knowledge!',
+    onRegister: () => alert('Registered!'),
+  },
+};
+
+export default meta;
```
Ditmar: This file exports `default meta` using a default export. Project conventions require named exports only. Storybook's CSF (component meta) typically uses a default export — if your project style rule forbids default exports globally, you must either add a per-file exception or alter the lint rule for stories.

**Sugerencia:**
```typescript
Option A (recommended practical): keep Storybook default export (required by Storybook) and add an inline ESLint exception comment for this line (e.g. `// eslint-disable-next-line import/no-default-export -- Storybook meta requires default export`).
Option B: if you strictly want no default exports, update project conventions to allow this pattern for *.stories.tsx files and document the exception in the coding guidelines.
```

src/components/UpcomingQuizCard/UpcomingQuizCard.styles.ts
```diff
@@ -0,0 +1,28 @@
...
+export const DateContainer = styled(Box)(({ theme }) => ({
+  margin: theme.spacing(2, 0),
+  display: 'flex',
+  flexDirection: 'column',
+  alignItems: 'center',
+}));
+
+export const RegisterButton = styled(Button)(({ theme }) => ({
+  marginTop: theme.spacing(2),
+  background: 'linear-gradient(90deg, #A44CE0, #6C63FF)',
```
Ditmar: Button styles include hard-coded colors and hex values (gradient and '#fff'). Project styling rules require using theme (palette/typography/spacing) instead of raw hex values to keep design tokens consistent and theme-aware.

**Sugerencia:**
```typescript
Use theme values. Example: use theme.palette.primary.main and theme.palette.secondary.main for gradients or define gradient tokens on the theme. Replace color '#fff' with theme.palette.primary.contrastText or theme.palette.common.white. Also avoid fractional spacing (theme.spacing expects numeric multiplier) — prefer integers or documented spacing values.

Example replacement:
background: `linear-gradient(90deg, ${theme.palette.primary.main}, ${theme.palette.secondary.main})`,
color: theme.palette.primary.contrastText,
padding: theme.spacing(1, 4),
```

src/components/UpcomingQuizCard/UpcomingQuizCard.test.tsx
```diff
@@ -0,0 +1,13 @@
+import { render, screen, fireEvent } from '@testing-library/react';
+import { describe, it, expect, vi } from 'vitest';
+
+import { UpcomingQuizCard } from './UpcomingQuizCard';
+
+describe('UpcomingQuizCard', () => {
+  it('calls onRegister when button is clicked', () => {
+    const handleRegister = vi.fn();
+    render(<UpcomingQuizCard title="Quiz" date="2023-08-12" onRegister={handleRegister} />);
+    fireEvent.click(screen.getByTestId('register-button'));
```
Ditmar: The test clicks the register button using getByTestId('register-button'). Prefer queries that reflect user behavior and accessibility, such as getByRole('button', { name: /register now/i }). Using accessible queries ensures your UI has the correct accessible name.

**Sugerencia:**
```typescript
Replace with an accessible query: `fireEvent.click(screen.getByRole('button', { name: /register now/i }));` Or use userEvent from @testing-library/user-event for more realistic interactions.
```

src/components/UpcomingQuizCard/UpcomingQuizCard.tsx
```diff
@@ -0,0 +1,58 @@
...
+export const UpcomingQuizCard: React.FC<UpcomingQuizCardProps> = ({
+  title,
+  date,
+  description,
+  onRegister,
+  ...cardProps
+}) => {
+  const { formattedDate } = useUpcomingQuizCard(date);
+
+  const getLocalDay = (dateValue: string | Date) => {
```
Ditmar: getLocalDay mixes manual string-splitting and UTC-based Date creation, then calls getUTCDate() for Date objects. This is inconsistent and error-prone. Day extraction should be consistent (either local or UTC) and ideally use a date parsing library to avoid timezone/parsing bugs.

**Sugerencia:**
```typescript
Use a single robust approach: parse ISO strings with date-fns parseISO and then use getDate() on the resulting Date (local day) or format(date, 'd') to get the day string. Example:

import { parseISO } from 'date-fns';
const parsed = typeof dateValue === 'string' ? parseISO(dateValue) : dateValue;
return parsed.getDate();

This keeps behavior consistent for strings and Date objects.
```

src/components/UpcomingQuizCard/UpcomingQuizCard.tsx
```diff
@@ -0,0 +1,58 @@
...
+        </DateContainer>
+
+        {description && (
+          <Typography variant="body2" color="text.secondary" gutterBottom>
+            {description}
+          </Typography>
+        )}
+
+        <RegisterButton variant="contained" onClick={onRegister} data-testid="register-button">
+          Register Now
```
Ditmar: The button label 'Register Now' is hard-coded in the component. Project guidelines require no hard-coded user-facing strings; they must be localizable.

**Sugerencia:**
```typescript
Accept the label as a prop (e.g., `registerLabel?: string`) or fetch the translated string via your i18n system (e.g., `t('upcomingQuiz.register')`) and avoid literal strings inside the component.
```

src/components/UpcomingQuizCard/UpcomingQuizCard.utils.ts
```diff
@@ -0,0 +1,10 @@
+import { format } from 'date-fns';
+
+import { DEFAULT_DATE_FORMAT } from './UpcomingQuizCard.constants';
+
+export const formatDate = (
+  date: string | Date,
+  formatStr: string = DEFAULT_DATE_FORMAT
+): string => {
+  return format(new Date(date), formatStr);
```
Ditmar: formatDate uses `new Date(date)` which can produce inconsistent results for ISO strings and does not validate the input. Prefer parseISO for YYYY-MM-DD and provide locale options when formatting.

**Sugerencia:**
```typescript
Use date-fns parseISO for ISO date strings and provide optional locale support. Example:

import { parseISO } from 'date-fns';
const d = typeof date === 'string' ? parseISO(date) : date;
return format(d, formatStr);

Also guard against invalid dates (isValid from date-fns) and return an empty string or fallback.
```

src/components/UpcomingQuizCard/icon/CalendarIcon.tsx
```diff
@@ -0,0 +1,58 @@
...
+    <Box component="span" sx={{ display: 'inline-flex', alignItems: 'center', ...sx }}>
+      <svg width={size} height={size} viewBox="0 0 64 64" xmlns="http://www.w3.org/2000/svg">
+        <rect
+          x="4"
+          y="8"
+          width="56"
+          height="52"
+          rx="8"
+          ry="8"
+          fill="#f0f0f0"
```
Ditmar: Several SVG fill/stroke values are hard-coded (e.g., `fill="#f0f0f0"`, `stroke="#ccc"`). These should be derived from the theme (palette) or use `currentColor` so that color can be controlled by parent components and theme.

**Sugerencia:**
```typescript
Use theme tokens or `currentColor` for SVG shapes. Example: `fill="currentColor"` and set color via the Box sx: `<Box sx={{ color: theme.palette.background.default }}>` or pass explicit theme-derived fills from parent.
```

src/components/UpcomingQuizCard/icon/CalendarIcon.tsx
```diff
@@ -0,0 +1,58 @@
...
+          strokeWidth="2"
+        />
+        <circle cx="20" cy="8" r="3" fill={color} />
+        <circle cx="44" cy="8" r="3" fill={color} />
+        <line x1="4" y1="20" x2="60" y2="20" stroke="#ccc" strokeWidth="2" />
+        <text
+          x="32"
+          y="46"
+          textAnchor="middle"
+          fontSize="24"
```
Ditmar: SVG <text> uses hard-coded fontSize="24". Font sizes should ideally rely on theme typography or relative units to remain consistent with the app theme.

**Sugerencia:**
```typescript
Use an explicit style that can come from theme, or set the font size via the parent `sx` prop. Example `fontSize: theme.typography.h6.fontSize` or use `style={{ fontSize: 'inherit' }}` and set font size via the wrapper.
```

src/components/UpcomingQuizCard/icon/CalendarIcon.tsx
```diff
@@ -0,0 +1,58 @@
...
+    small: 32,
+    medium: 48,
+    large: 64,
+  };
+
+  const size = sizeMap[fontSize] || 48;
+
+  return (
+    <Box component="span" sx={{ display: 'inline-flex', alignItems: 'center', ...sx }}>
+      <svg width={size} height={size} viewBox="0 0 64 64" xmlns="http://www.w3.org/2000/svg">
```
Ditmar: SVG element does not provide an accessible name or role. If the icon conveys meaningful information (the day), it should be accessible (role and aria-label or visually-hidden text).

**Sugerencia:**
```typescript
Add role and accessible name when appropriate: `<svg role="img" aria-label={`Calendar day ${day}`} ...>` or allow passing an `aria-label` prop. Alternatively, if the calendar icon is purely decorative, add `aria-hidden="true"`.
```

Ditmar (changes requested):
The feature is on the right track and mostly follows CDD structure, but there are critical functional and convention issues to resolve before approval. Highest-priority fixes: remove the +1 day logic in the hook (causes incorrect dates), convert default exports to named exports (or document a Storybook exception), and replace hard-coded color/typography tokens with theme-driven values. Also improve parsing with date-fns parseISO and make strings localizable. After addressing these points (and updating tests to use accessible queries), re-run lint/build/tests and submit an updated PR for final review.

Trocador:
Try to add a capture of the test results in the terminal and repair the table in the first docs:
image
i think is defased the columns

Ditmar (approved):

Ditmar (dismissed):

PR #77, opened 2025-10-12
[CDD] Build AvatarMenu component

Adrex-2003:
# Initial Implementation of AvatarMenu Component

An accessible dropdown menu for authenticated users

image
image

## Component Purpose
Provides a centralized access point for user actions and settings within the application, with special focus on:

- Complete accessibility (ARIA, keyboard navigation)
- Intuitive user experience
- Seamless integration with Material-UI
- Responsive behavior

### Core Functionality
- Dropdown menu with user avatar and username
- Customizable items with icons, dividers, and disabled states
- Complete keyboard navigation (Enter, Escape, arrow keys)
- Intelligent zoom behavior (automatic close)
- FullWidth support for responsive layouts

### Accessibility
- Implemented ARIA attributes (`aria-haspopup`, `aria-expanded`, `aria-controls`)
- Complete keyboard navigation
- Focus management and visual indicators
- Screen reader support

### UI/UX
- Consistent design system with Material-UI
- Clear visual states (hover, focus, active)
- Smooth transitions and immediate feedback
- Animated dropdown indicator icon

### Storybook
https://ditmar.github.io/leetcode-spa/feature-leetcode-19/

### DIFF
--- a/src/components/AvatarMenu/AvatarMenu.styles.ts
+++ b/src/components/AvatarMenu/AvatarMenu.styles.ts
@@ -1,17 +1,23 @@
 import { IconButton, Typography } from '@mui/material';
 import { styled } from '@mui/material/styles';
 
+import { theme } from '../../style-library';
+
 import type { StyledAvatarMenuProps } from './AvatarMenu.types';
 
 export const AvatarMenuRoot = styled(IconButton, {
   shouldForwardProp: (prop) => prop !== '$fullWidth' && prop !== '$isOpen',
 })<StyledAvatarMenuProps>((props) => {
+  const isDarkMode = (theme.palette.mode = 'dark');
   return {
     borderRadius: props.theme.shape.borderRadius,
     padding: props.theme.spacing(1),
     width: props.$fullWidth ? '100%' : 'auto',
     justifyContent: props.$fullWidth ? 'flex-end' : 'center',
     display: 'flex',
+    '&:hover': {
+      backgroundColor: isDarkMode ? theme.palette.action.focus : theme.palette.action.hover,
+    },
   };
 });
 
@@ -20,7 +26,7 @@ export const AvatarContainer = styled('div')<{
 }>(({ theme, $fullWidth }) => ({
   display: 'flex',
   alignItems: 'center',
-  gap: theme.spacing(1.025),
+  gap: theme.spacing(1.125),
   position: 'relative',
   width: $fullWidth ? '100%' : 'auto',
   justifyContent: $fullWidth ? 'flex-end' : 'center',
@@ -34,13 +40,20 @@ export const UsernameText = styled(Typography)(({ theme }) => ({
   },
 }));
 
+export const ArrowWrapper = styled('div')(({ theme }) => ({
+  width: theme.spacing(2),
+  height: theme.spacing(2),
+  display: 'flex',
+  alignItems: 'center',
+  justifyContent: 'center',
+}));
+
 export const TriangleIndicator = styled('div')<{ $isOpen: boolean }>(({ theme, $isOpen }) => ({
   width: 0,
   height: 0,
   borderTop: `${theme.spacing(1.25)} solid ${theme.palette.text.primary}`,
   borderLeft: `${theme.spacing(0.833)} solid transparent`,
   borderRight: `${theme.spacing(0.833)} solid transparent`,
-  marginLeft: theme.spacing(0.5),
   transition: theme.transitions.create(['transform'], {
     duration: theme.transitions.duration.short,
   }),
--- a/src/components/AvatarMenu/AvatarMenu.tsx
+++ b/src/components/AvatarMenu/AvatarMenu.tsx
@@ -4,7 +4,12 @@ import React from 'react';
 
 import { DEFAULT_AVATAR_MENU_ITEMS } from './AvatarMenu.constants';
 import { useAvatarMenu } from './AvatarMenu.hook';
-import { AvatarMenuRoot, AvatarContainer, TriangleIndicator } from './AvatarMenu.styles';
+import {
+  AvatarMenuRoot,
+  AvatarContainer,
+  TriangleIndicator,
+  ArrowWrapper,
+} from './AvatarMenu.styles';
 
 import type { AvatarMenuProps } from './AvatarMenu.types';
 
@@ -57,7 +62,9 @@ export const AvatarMenu: React.FC<AvatarMenuProps> = ({
             sx={{ ...avatarBaseSx, ...customSx }}
             {...restAvatarProps}
           />
-          <TriangleIndicator $isOpen={isOpen} />
+          <ArrowWrapper>
+            <TriangleIndicator $isOpen={isOpen} />
+          </ArrowWrapper>
         </AvatarContainer>
       </AvatarMenuRoot>
 
--- a/yarn.lock
+++ b/yarn.lock
@@ -473,135 +473,135 @@
   resolved "https://registry.yarnpkg.com/@emotion/weak-memoize/-/weak-memoize-0.4.0.tgz#5e13fac887f08c44f76b0ccaf3370eb00fec9bb6"
   integrity sha512-snKqtPW01tN0ui7yu9rGv69aJXr/a/Ywvl11sUjNtEcRc+ng/mQriFL0wLXMef74iHa/EkftbDzU9F8iFbH+zg==
 
-"@esbuild/aix-ppc64@0.25.10":
-  version "0.25.10"
-  resolved "https://registry.yarnpkg.com/@esbuild/aix-ppc64/-/aix-ppc64-0.25.10.tgz#ee6b7163a13528e099ecf562b972f2bcebe0aa97"
-  integrity sha512-0NFWnA+7l41irNuaSVlLfgNT12caWJVLzp5eAVhZ0z1qpxbockccEt3s+149rE64VUI3Ml2zt8Nv5JVc4QXTsw==
+"@esbuild/aix-ppc64@0.25.11":
+  version "0.25.11"
+  resolved "https://registry.yarnpkg.com/@esbuild/aix-ppc64/-/aix-ppc64-0.25.11.tgz#2ae33300598132cc4cf580dbbb28d30fed3c5c49"
+  integrity sha512-Xt1dOL13m8u0WE8iplx9Ibbm+hFAO0GsU2P34UNoDGvZYkY8ifSiy6Zuc1lYxfG7svWE2fzqCUmFp5HCn51gJg==
 
-"@esbuild/android-arm64@0.25.10":
-  version "0.25.10"
-  resolved "https://registry.yarnpkg.com/@esbuild/android-arm64/-/android-arm64-0.25.10.tgz#115fc76631e82dd06811bfaf2db0d4979c16e2cb"
-  integrity sha512-LSQa7eDahypv/VO6WKohZGPSJDq5OVOo3UoFR1E4t4Gj1W7zEQMUhI+lo81H+DtB+kP+tDgBp+M4oNCwp6kffg==
+"@esbuild/android-arm64@0.25.11":
+  version "0.25.11"
+  resolved "https://registry.yarnpkg.com/@esbuild/android-arm64/-/android-arm64-0.25.11.tgz#927708b3db5d739d6cb7709136924cc81bec9b03"
+  integrity sha512-9slpyFBc4FPPz48+f6jyiXOx/Y4v34TUeDDXJpZqAWQn/08lKGeD8aDp9TMn9jDz2CiEuHwfhRmGBvpnd/PWIQ==
 
-"@esbuild/android-arm@0.25.10":
-  version "0.25.10"
-  resolved "https://registry.yarnpkg.com/@esbuild/android-arm/-/android-arm-0.25.10.tgz#8d5811912da77f615398611e5bbc1333fe321aa9"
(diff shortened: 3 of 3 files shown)


src/components/AvatarMenu/AvatarMenu.stories.tsx
```diff
@@ -0,0 +1,140 @@
...
+const meta: Meta<AvatarMenuStoryArgs> = {
+	title: 'Navigation/AvatarMenu',
+	component: AvatarMenu,
+	parameters: {
+		layout: 'centered',
+		docs: {
+			description: {
+				component: 'Componente de menú desplegable para usuario autenticado',
+			},
+		},
```
Ditmar: You use an arg named 'data-testid' (with a hyphen) in argTypes/args. Storybook args with hyphenated names are awkward and can cause problems; prefer camelCase prop names in stories/controls.

**Sugerencia:**
```typescript
Change story arg keys to camelCase (e.g. dataTestId) and map them to component props if necessary. This keeps controls simpler and consistent.
```

src/components/AvatarMenu/AvatarMenu.styles.ts
```diff
@@ -0,0 +1,51 @@
+import { styled } from '@mui/material/styles';
+import { IconButton, Typography } from '@mui/material';
+import type { PropsStyledAvatarMenu } from './AvatarMenu.types';
+import { theme } from '../../style-library';
+
```
Ditmar: This file imports a top-level `theme` from '../../style-library' and then uses it directly inside styled() callbacks. Mixing the imported theme with the styled-provided theme is fragile and breaks theming (and SSR).

**Sugerencia:**
```typescript
Remove the imported theme and use the theme provided by MUI inside the styled callbacks: e.g. styled('div')(({ theme, $fullWidth }) => ({ gap: theme.spacing(1.25) }))
```

src/components/AvatarMenu/AvatarMenu.styles.ts
```diff
@@ -0,0 +1,51 @@
...
+export const AvatarMenuRoot = styled(IconButton, {
+	shouldForwardProp: (prop) => prop !== '$fullWidth',
+})<PropsStyledAvatarMenu>((props) => {
```
Ditmar: shouldForwardProp excludes only '$fullWidth' but not '$isOpen'. In AvatarMenu you pass $isOpen to AvatarMenuRoot (see AvatarMenu.tsx line 43). Because $isOpen is not filtered it will be forwarded to the DOM producing React warnings for unknown attributes.

**Sugerencia:**
```typescript
Exclude all $-prefixed styling-only props from being forwarded. Example:
shouldForwardProp: (prop) => prop !== '$fullWidth' && prop !== '$isOpen'
```

src/components/AvatarMenu/AvatarMenu.styles.ts
```diff
@@ -0,0 +1,51 @@
...
+});
+
+
+export const AvatarContainer = styled('div')<{
+	$fullWidth?: boolean;
+}>(({ $fullWidth }) => ({
+	display: 'flex',
+	alignItems: 'center',
+	gap: theme.spacing(1.25),
+	position: 'relative',
```
Ditmar: AvatarContainer uses the imported theme variable rather than the theme provided by the styled callback. Also you typed $fullWidth as optional but AvatarMenuRoot typed PropsStyledAvatarMenu requires $fullWidth: boolean — there's an inconsistency between these types/usages.

**Sugerencia:**
```typescript
Use: styled('div')<Props>(({ theme, $fullWidth }) => ({ gap: theme.spacing(1.25), ... })) and align PropsStyledAvatarMenu to allow optional $fullWidth if that's the intention.
```

src/components/AvatarMenu/AvatarMenu.styles.ts
```diff
@@ -0,0 +1,51 @@
...
+	[theme.breakpoints.down('sm')]: {
+		display: 'none',
+	},
+}));
+
+export const TriangleIndicator = styled('div')<{ $isOpen: boolean }>(({ theme, $isOpen }) => ({
+	width: 0,
+	height: 0,
+	borderLeft: '6px solid transparent',
+	borderRight: '6px solid transparent',
```
Ditmar: TriangleIndicator uses hardcoded pixel values for borders and sizes (e.g. '6px', '8px'). Project rules require avoiding hardcoded px values and using theme spacing / tokens.

**Sugerencia:**
```typescript
Use theme-based spacing and sizes. Example:
borderLeft: `${theme.spacing(0.75)} solid transparent`,
borderTop: `${theme.spacing(1)} solid ${theme.palette.text.primary}`
(or compute equivalent values using theme.spacing).
```

src/components/AvatarMenu/AvatarMenu.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+};
+
+describe('AvatarMenu', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+  const getAvatarButton = (username: string = mockProps.username) => {
+    return screen.getByRole('button', {
+      name: `Menú de usuario${username ? `: ${username}` : ''}`
+    });
```
Ditmar: Tests use a Spanish aria-label check to find the button (getByRole('button', { name: `Menú de usuario...` })). The project mandates English as the codebase standard; tests and component labels should be in English. Also relying on exact aria-label strings in multiple tests couples tests tightly to text that may change (i18n).

**Sugerencia:**
```typescript
Change component aria-label to English (e.g. 'User menu') and update tests accordingly. Prefer using aria-label constants or data-testid for stable targeting, and prefer role-based selectors like getByRole('button', { name: /user menu/i }) for resilience.
```

src/components/AvatarMenu/AvatarMenu.tsx
```diff
@@ -0,0 +1,131 @@
...
+				}}
+				transformOrigin={{
+					vertical: 'top',
+					horizontal: 'right',
+				}}
+				slotProps={{
+					paper: {
+						sx: {
+							marginTop: '8px',
+							zIndex: 1300,
```
Ditmar: You set marginTop: '8px' inside slotProps.paper.sx which is a hardcoded px value. Per styling rules, avoid raw px and prefer theme spacing tokens.

**Sugerencia:**
```typescript
Use theme spacing via sx callback, e.g. paper: { sx: (theme) => ({ mt: theme.spacing(1), zIndex: theme.zIndex.modal }) } or marginTop: theme.spacing(1).
```

src/components/AvatarMenu/AvatarMenu.tsx
```diff
@@ -0,0 +1,131 @@
...
+					dense: true,
+				}}
+			>
+				{menuItems.map((item, index) => {
+					const isEnabled = !item.disabled;
+					const isFocused = focusedIndex === index && isEnabled;
+
+					return (
+						<div key={item.label || index}>
+							<MenuItem
```
Ditmar: You wrap each MenuItem (and optional Divider) inside a plain <div> with a key. Wrapping MenuItem in an extra div can break keyboard navigation and semantics of Menu. Also using item.label as the key may not be unique.

**Sugerencia:**
```typescript
Use React.Fragment for grouping and place the key on the MenuItem (or use a stable unique id). Example:
return (
  <Fragment key={`${item.label}-${index}`}>
    <MenuItem ... />
    {item.divider && <Divider />}
  </Fragment>
);
Or set key directly on MenuItem: <MenuItem key={`menuitem-${index}`}>
```

src/components/AvatarMenu/AvatarMenu.tsx
```diff
@@ -0,0 +1,131 @@
...
+										setFocusedIndex(index);
+									}
+								}}
+								disabled={item.disabled}
+								data-testid={item['data-testid']}
+								autoFocus={isFocused}
+								{...item.menuItemProps}
+								sx={{
+									...(isFocused && {
+										backgroundColor: 'action.hover',
```
Ditmar: You set borderRadius and other style attributes in sx with hardcoded pixel values (borderRadius: '6px') and boxShadow string. Project guidelines recommend using theme tokens and avoiding px hardcodes.

**Sugerencia:**
```typescript
Use theme tokens in sx (e.g. borderRadius: 1 or theme.shape.borderRadius) and theme shadows (theme.shadows[1] or sx: { boxShadow: (theme) => theme.shadows[1] }). Example:
...(isFocused && {
  backgroundColor: 'action.hover',
  boxShadow: (theme) => theme.shadows[1],
  borderRadius: (theme) => theme.shape.borderRadius,
}),
```

src/components/AvatarMenu/AvatarMenu.types.ts
```diff
@@ -0,0 +1,27 @@
...
+	avatarUrl: string;
+	username: string;
+	menuItems?: AvatarMenuItem[];
+	avatarProps?: Omit<MuiAvatarProps, 'src' | 'alt'>;
+	fullWidth?: boolean;
+	'data-testid'?: string;
+}
+
+export interface PropsStyledAvatarMenu {
+	$isOpen: boolean;
```
Ditmar: PropsStyledAvatarMenu defines both $isOpen and $fullWidth as required booleans. In usage $fullWidth can be optional; consider making these props optional (or ensure you always pass both). Also styled props should be prefixed with $ and omitted from DOM via shouldForwardProp.

**Sugerencia:**
```typescript
Make these props optional if not always provided: export interface PropsStyledAvatarMenu { $isOpen?: boolean; $fullWidth?: boolean; } and ensure shouldForwardProp excludes them all.
```

Ditmar (changes requested):
Overall this PR adds a new AvatarMenu component with stories, docs, tests, styles and an internal hook, and makes several package.json/yarn.lock changes. I found multiple functional, accessibility, typing and code-style issues that must be addressed before approval. Main categories: 1) English-only rule broken (Spanish used in code, docs and tests), 2) accessibility / focus management bugs (mismatch between enabled items count and focused index, wrapping MenuItem in a DIV), 3) styling API misuse and leaking props to DOM ($isOpen forwarded), 4) unsafe window usage and incorrect event typing in the hook (SSR and 'any' usage), 5) fragile story & storybook dependency mismatches, and 6) tests that can be flaky or rely on Spanish aria-labels and non-robust selectors. Read the comments below for line-by-line issues and suggested fixes.

### ℹ️ Comentarios Adicionales

Se identificaron **5 comentarios adicionales** en líneas que no están en el diff actual. Estos comentarios se incluyen aquí como sugerencias generales:

**1. package.json:14**
You added "@storybook/blocks": "8.6.14" while most other Storybook packages in this file are 9.1.10 (see lines 11-16). Mixing major versions (8.x and 9.x) can cause runtime and build incompatibilities.
*Sugerencia:* Align block package with the rest of Storybook (use a 9.x version) or remove it if not needed. Example:
"@storybook/blocks": "9.1.10",
Or remove the line entirely if blocks is not required by your Storybook config.

**2. package.json:15**
You added "@storybook/react": "9.1.10" while the project already includes "@storybook/react-vite": "9.1.10" (line 16). Ensure you actually need both packages; some setups only need the builder package (react-vite) and the top-level storybook packages can be redundant.
*Sugerencia:* If using Vite builder, you likely can remove the direct @storybook/react entry. If you need it, ensure consistency and deduplicate entries in package.json and yarn.lock.

**3. package.json:17**
You removed the caret prefix from @testing-library/jest-dom and @testing-library/react (changed to pinned versions). This is consistent with a policy to pin deps, but please confirm that's intentional (it prevents automatic minor/patch updates).
*Sugerencia:* If pinning is intentional, it's fine. Otherwise revert to using the caret if you want to allow minor updates, e.g.: "@testing-library/jest-dom": "^6.9.1".

**4. package.json:25**
You removed the caret from @types/react-dom (changed to "19.2.0"). Pinning types can be fine but be consistent across packages and team expectations.
*Sugerencia:* Ensure all @types packages follow your chosen versioning policy. If you pin type packages, verify compatibility with the installed react-dom version.

**5. src/components/AvatarMenu/AvatarMenu.constants.ts:21**
You used `as const` on the array while also explicitly typing the variable as AvatarMenuItem[]. `as const` produces a readonly tuple type which can be surprising when the variable is intended to be a mutable/regular array type.
*Sugerencia:* Either remove `as const` or change the exported type to readonly if you intend immutability. Example to keep simple:
export const DEFAULT_AVATAR_MENU_ITEMS: AvatarMenuItem[] = [ ... ];
Or, if immutability is desired:
export const DEFAULT_AVATAR_MENU_ITEMS = [ ... ] as const;
export type DefaultAvatarMenuItems = typeof DEFAULT_AVATAR_MENU_ITEMS;


---

**Conclusión:** Requesting changes. The new AvatarMenu implementation is a good start, but there are multiple important issues to fix: translate all user-facing strings/docs/tests to English, prevent styled-prop leakage ($isOpen), fix keyboard/focus logic so indexes align with rendered items, avoid direct window access without guards (SSR-safe), remove brittle label-based logic in stories, and adjust tests to use robust selectors. Address all line-level comments and then request another review.

package.json
```diff
@@ -39,16 +39,16 @@
     "@chromatic-com/storybook": "4.1.1",
     "@eslint/js": "9.36.0",
     "@storybook/addon-a11y": "9.1.10",
-    "@storybook/addon-docs": "9.1.10",
+    "@storybook/addon-docs": "^9.1.10",
```
Ditmar: borra el carret (^)

Ditmar (approved):

Adrex-2003:
## Storybook
https://ditmar.github.io/leetcode-spa/feature-leetcode-19/

Ditmar (dismissed):

RosaTorrez (changes requested):
#109

Ditmar (approved):

RosaTorrez (approved):
#109

Issue #22, opened 2025-10-01
[CDD ] Create FormInput component


Ditmar:
Implement a FormInput component representing a styled input field for forms (e.g., "Mail Id", "Username") based on the design.

**Requirements:**
- Accept props such as `label`, `placeholder`, `value`, `type` (text/email/password), `onChange`, `disabled`, and `errorMessage`.
- Render a MUI TextField with custom styles using `styled()` or `sx` and theme spacing/color tokens; support variations (outlined/filled) and show error state when `errorMessage` is provided.
- Provide accessibility attributes (e.g., `id`, `aria-label`) and support passing additional attributes to the underlying input.
- Manage focus state and optionally call `onFocus`/`onBlur` if provided; isolate state management in a hook if necessary to keep the component clean.
- Extend MUI input props via a `FormInputProps` interface and define constants/enums for variant or size options where needed.
- Include unit tests validating that the component renders the label, passes value and placeholder correctly, triggers the `onChange` callback, and shows error state when `errorMessage` is set.
- Create Storybook stories demonstrating default, filled, disabled, and error states; include interactive controls for `label`, `placeholder`, and `type`.
- Provide MDX documentation explaining usage, supported props, and integration with form containers.

### File structure
FormInput/
  ├── FormInput.tsx
  ├── FormInput.types.ts
  ├── FormInput.constants.ts (optional)
  ├── FormInput.styles.ts
  ├── FormInput.hook.ts (optional)
  ├── FormInput.utils.ts (optional)
  ├── FormInput.test.tsx
  ├── FormInput.stories.tsx
  └── FormInput.docs.mdx

PR #91, opened 2025-10-14
[CDD] Create FormInput component

JoseHuanaco:
Implementation of the FormInput component as per ticket requirements, providing a styled input field for forms with support for variants, sizes, states, and accessibility.

Resolves: #22 

Implementation:

- Created FormInput component extending MUI TextField with custom styling using styled() and theme tokens.
- Implemented support for placeholder, value, type (text/email/password), onChange, disabled, and errorMessage props.
- Added FormInput.types.ts with FormInputProps interface extending MUI types.
- Defined constants in FormInput.constants.ts for variants, sizes, types, and color tokens.
- Included 7 unit tests in FormInput.test.tsx validating rendering, value/placeholder handling, onChange callback, and error states.
- Created Storybook stories (Playground, Default, Filled, Disabled, Error) with interactive controls.
- Added FormInput.docs.mdx documentation explaining usage.

image

image

image

image

image

image

### DIFF
--- a/src/components/FormInput/FormInput.styles.ts
+++ b/src/components/FormInput/FormInput.styles.ts
@@ -39,12 +39,12 @@ export const StyledFormInput = styled(TextField, {
   // Function for eye icon sizes
   const getIconStyles = (size: 'small' | 'medium' | 'large') => ({
     '& .MuiIconButton-root': {
-      width: `${formInputTokens.eye.size[size].width}px !important`,
-      height: `${formInputTokens.eye.size[size].height}px !important`,
+      width: `${formInputTokens.eye.sizeBox[size].width}px !important`,
+      height: `${formInputTokens.eye.sizeBox[size].height}px !important`,
       padding: `${formInputTokens.eye.padding[size]}px !important`,
       '& .MuiSvgIcon-root': {
-        width: `${formInputTokens.eye.size[size].width}px !important`,
-        height: `${formInputTokens.eye.size[size].height}px !important`,
+        width: `${formInputTokens.eye.sizeEye[size].width}px !important`,
+        height: `${formInputTokens.eye.sizeEye[size].height}px !important`,
       },
     },
   });
--- a/src/style-library/theme/theme.ts
+++ b/src/style-library/theme/theme.ts
@@ -150,15 +150,20 @@ export const formInputTokens = {
     },
   },
   eye: {
-    size: {
+    sizeBox: {
       small: { width: 15.34, height: 15.16 },
       medium: { width: 18.5, height: 18.5 },
+      large: { width: 21.93, height: 21.93 },
+    },
+    sizeEye: {
+      small: { width: 13.74, height: 11.06 },
+      medium: { width: 16.6, height: 13.55 },
       large: { width: 19.64, height: 15.99 },
     },
     padding: {
-      small: 10,
-      medium: 14,
-      large: 18,
+      small: 0,
+      medium: 0,
+      large: 0,
     },
   },
 } as const;


src/components/FormInput/FormInput.styles.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export const StyledFormInput = styled(TextField, {
+  shouldForwardProp: (prop) => (typeof prop === 'string' ? !prop.startsWith('$') : true),
+})<StyledFormInputProps>(({ theme, $variant, $size, $error }) => {
+  const sizeConfig = $size ? SIZE_CONFIGS[$size] : SIZE_CONFIGS['medium'];
+
+  return {
+    width: '100%',
+    '& .MuiInputBase-root': {
+      height: theme.spacing(Number(sizeConfig.height.replace('rem', '')) * 2),
```
Ditmar: The height is computed by parsing a rem string and feeding an ad-hoc Number(...) * 2 into theme.spacing: height: theme.spacing(Number(sizeConfig.height.replace('rem', '')) * 2). This is brittle (string manipulation), incorrect semantics (theme.spacing expects a spacing factor, not a rem), and will break if SIZE_CONFIGS uses different units.

**Sugerencia:**
```typescript
Use numeric tokens in SIZE_CONFIGS (see suggestions in constants file) and compute height consistently. Example using numeric rem tokens and converting to px/rem properly:

const heightRem = sizeConfig.heightRem; // number
height: `${heightRem}rem`,

Or derive using theme.spacing if you store spacing units: height: theme.spacing(heightUnits).

Avoid parsing strings in styles.
```

src/components/FormInput/FormInput.styles.ts
```diff
@@ -0,0 +1,69 @@
...
+      },
+    },
+    '& .MuiOutlinedInput-root': {
+      '& fieldset': {
+        borderColor: $error ? theme.palette.error.main : theme.palette.divider,
+        borderRadius: theme.shape.borderRadius,
+        borderWidth: '1px',
+      },
+      '&:hover fieldset': {
+        borderColor: $error ? theme.palette.error.main : theme.palette.custom?.focus,
```
Ditmar: The style references theme.palette.custom?.focus. 'palette.custom' is not a standard MUI palette key and may be undefined at runtime or break theme typings unless the theme augmentation is present. Optional chaining prevents runtime crash but it hides the fact that the design token may not exist.

**Sugerencia:**
```typescript
Use a palette key that exists on your theme (e.g., primary.main, secondary.main) or ensure theme augmentation types define palette.custom.focus. If you rely on a custom token, add a fallback value, e.g.:

borderColor: $error ? theme.palette.error.main : (theme.palette.custom?.focus ?? theme.palette.primary.main),

and make sure to update theme types accordingly.
```

src/components/FormInput/FormInput.styles.ts
```diff
@@ -0,0 +1,69 @@
...
+        borderWidth: '1px',
+      },
+      '&.Mui-focused fieldset': {
+        borderColor: theme.palette.custom?.focus,
+        borderWidth: '1px',
+      },
+      ...($variant === 'filled' && {
```
Ditmar: Using a computed spread with ...($variant === 'filled' && { ... }) will spread 'false' when $variant !== 'filled' which actually results in no-op, but the pattern is harder to read and may confuse. Additionally, inside you remove the fieldset border (border: 'none') while leaving other selectors intact. Verify that for the 'filled' variant the MUI classnames used here are correct and consistent.

```typescript
Prefer a clear conditional object merge or ternary to return an object. Example:

...($variant === 'filled' ? {
  '& fieldset': { border: 'none', backgroundColor: theme.palette.action.selected }
} : {}),

Also confirm the correct MUI classes for 'filled' variant.
```

src/components/FormInput/FormInput.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const handleFocus = (event: React.FocusEvent<HTMLInputElement>) => {
+    if (onFocus) onFocus(event);
+  };
+
+  const handleBlur = (event: React.FocusEvent<HTMLInputElement>) => {
+    if (onBlur) onBlur(event);
+  };
+
+  return (
+    <StyledFormInput
```
Ditmar: The component passes props to StyledFormInput using standard prop names (variant, size, error) while the styled component expects $variant, $size, $error (see styles). This mismatch prevents the styled layer from receiving the intended values and may lead to styling not being applied.

**Sugerencia:**
```typescript
Pass the $-prefixed props to the styled component in addition to the original props if needed by MUI. Example:

<StyledFormInput
  $variant={variant}
  $size={size}
  $error={isError}
  variant={variant}
  size={size}
  error={isError}
  ...
/>

This keeps MUI props intact and supplies the styled component with the values it expects.
```

src/components/FormInput/FormInput.tsx
```diff
@@ -0,0 +1,52 @@
...
+      placeholder={placeholder}
+      value={value}
+      onChange={onChange}
+      error={isError}
+      helperText={isError ? errorMessage : undefined}
+      disabled={disabled}
+      onFocus={handleFocus}
+      onBlur={handleBlur}
+      data-testid={dataTestId}
+      {...rest}
```
Ditmar: You spread {...rest} after setting data-testid and before inputProps. If the caller passes inputProps via rest, your inline inputProps will override theirs (or be overridden depending on order). Also inputProps is being provided as a literal which may clobber any user-provided inputProps. This can unintentionally discard attributes.

**Sugerencia:**
```typescript
Merge inputProps safely. Example:

const mergedInputProps = { ...(rest.inputProps || {}), 'aria-label': placeholder || 'input' };

Then pass inputProps={mergedInputProps} and spread rest without inputProps (or remove inputProps from rest when spreading):

const { inputProps: _unused, ...restWithoutInputProps } = rest;

<StyledFormInput {...restWithoutInputProps} inputProps={mergedInputProps} />

This ensures callers can pass additional inputProps while preserving aria-label.
```

src/components/FormInput/FormInput.tsx
```diff
@@ -0,0 +1,52 @@
...
+      inputProps={{ 'aria-label': placeholder || 'input' }}
+    />
+  );
+};
+
+export default FormInput;
```
Ditmar: This file exports the component as a default export (export default FormInput). The project rules state that named exports must be used and default exports are forbidden.

**Sugerencia:**
```typescript
Remove the default export and only use the named export already present. For example, delete line 52. Consumers should import with: import { FormInput } from './FormInput';
```

src/components/FormInput/FormInput.types.ts
```diff
@@ -0,0 +1,26 @@
+import type { TextFieldProps } from '@mui/material/TextField';
```
Ditmar: This file references React types in the props signatures (React.ChangeEvent, React.FocusEvent) but does not import React or the specific event types. Depending on tsconfig settings, this may cause 'React' is not defined errors in a types-only file.

**Sugerencia:**
```typescript
Add an explicit type-only import. Example:

import type React from 'react';

or import specific types:

import type { ChangeEvent, FocusEvent } from 'react';

Then use ChangeEvent<HTMLInputElement> / FocusEvent<HTMLInputElement> in your prop types to avoid referencing the namespace React directly.
```

src/components/FormInput/FormInput.types.ts
```diff
@@ -0,0 +1,26 @@
...
+  onFocus?: (event: React.FocusEvent<HTMLInputElement>) => void;
+  onBlur?: (event: React.FocusEvent<HTMLInputElement>) => void;
+  'data-testid'?: string;
+}
+
+export interface StyledFormInputProps {
+  $variant: FormInputVariant;
+  $size: FormInputSize;
+  $error: boolean;
+  $shape?: keyof typeof SHAPE_CONFIGS;
```
Ditmar: StyledFormInputProps references SHAPE_CONFIGS in the $shape property (keyof typeof SHAPE_CONFIGS) but SHAPE_CONFIGS is not defined in this file or in the diffs. This creates a missing symbol and will cause a type error.

**Sugerencia:**
```typescript
Either define SHAPE_CONFIGS in this file (or import it from the constants file) or remove the $shape prop if it's not implemented yet. Example:

// if shape support is intended
export const SHAPE_CONFIGS = { /* ... */ } as const;

or change the type to: $shape?: string; // or a union of allowed shapes
```

src/components/FormInput/FormInput.test.tsx
```diff
@@ -0,0 +1,88 @@
+import userEvent from '@testing-library/user-event';
+import { useState } from 'react';
+import { describe, it, expect, vi } from 'vitest';
+
+import { render, screen } from '../../test/test-utils';
+
+import { FormInput } from './FormInput';
+
+describe('FormInput', () => {
+  const defaultProps = {
```
Ditmar: defaultProps contains a single shared mock for onChange (vi.fn()). Reusing the same mock across multiple tests without resetting it can cause cross-test leakage and false positives/failures.

**Sugerencia:**
```typescript
Create fresh mocks per test or reset mocks in a beforeEach. Example:

let onChange: ReturnType<typeof vi.fn>;
beforeEach(() => { onChange = vi.fn(); });
// use onChange in the render for each test

This ensures test isolation.
```

src/components/FormInput/FormInput.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+    label: 'Mail Id',
+    value: '',
+    onChange: vi.fn(),
```
Ditmar: defaultProps defines 'onChange: vi.fn()' at module scope. This common mock will accumulate calls over the test suite unless reset. See above suggestion to create or reset mocks per test.

**Sugerencia:**
```typescript
Move vi.fn() creation into individual tests or a beforeEach and reference that mock inside each test rather than sharing a single mock instance globally.
```

src/components/FormInput/FormInput.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+    'data-testid': 'form-input',
+  };
+
+  const FormInputWrapper = ({ value: initialValue = '', onChange, ...props }) => {
```
Ditmar: FormInputWrapper is an untyped React component (props are not typed). The project forbids implicit any and prefers explicit types for component props. Also it defines an internal state for value — good for testing the controlled input, but types should be explicit.

**Sugerencia:**
```typescript
Add a props type. Example:

interface FormInputWrapperProps extends Partial<Pick<FormInputProps, 'value' | 'onChange'>> {
  [key: string]: any;
}

const FormInputWrapper: React.FC<FormInputWrapperProps> = ({ value: initialValue = '', onChange, ...props }) => { ... }

Or import FormInputProps and use Omit to type allowed props for the wrapper.
```

src/components/FormInput/FormInput.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+    const [value, setValue] = useState(initialValue);
+    const handleChange = (e) => {
+      setValue(e.target.value);
+      if (onChange) onChange(e);
+    };
+    return <FormInput {...props} value={value} onChange={handleChange} />;
+  };
+
+  it('renderiza con props por defecto', () => {
```
Ditmar: The test description strings are in Spanish (e.g. 'renderiza con props por defecto'). The project requires English as the standard for code, comments, and tests.

**Sugerencia:**
```typescript
Translate test descriptions to English. Example:

it('renders with default props', () => { ... });
```

src/components/FormInput/FormInput.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+    render(<FormInputWrapper {...defaultProps} />);
+    const inputContainer = screen.getByTestId('form-input');
+    expect(inputContainer).toBeInTheDocument();
+    expect(screen.getByLabelText('Mail Id')).toBeInTheDocument();
+    expect(screen.getByRole('textbox')).toHaveValue('');
+  });
+
+  it('renderiza con placeholder y value correctos', () => {
```
Ditmar: Another test description is in Spanish ('renderiza con placeholder y value correctos'). Tests should use English descriptions per project rules.

**Sugerencia:**
```typescript
Translate to English, eg: it('renders with correct placeholder and value', () => { ... });
```

src/components/FormInput/FormInput.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+    await userEvent.type(input, 'test@example.com');
+    expect(input).toHaveValue('test@example.com');
+    expect(defaultProps.onChange).toHaveBeenCalledWith(
+      expect.objectContaining({
+        target: expect.objectContaining({ value: 'test@example.com' }),
+      })
+    );
+  });
+
+  it('no dispara onChange si está disabled', async () => {
```
Ditmar: Test description in Spanish ('no dispara onChange si está disabled'). Use English and ensure the test asserts behavior robustly.

**Sugerencia:**
```typescript
Translate to English: it('does not call onChange when disabled', async () => { ... }); Also consider verifying that the handler has not been called across the suite by using isolated mocks per test.
```

src/components/FormInput/FormInput.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+    const handleChange = vi.fn();
+    render(<FormInputWrapper {...defaultProps} onChange={handleChange} disabled />);
+    const input = screen.getByRole('textbox');
+    expect(input).toBeDisabled();
+    await userEvent.type(input, 'test');
+    expect(handleChange).not.toHaveBeenCalled();
+  });
+
+  it('muestra estado de error con errorMessage', () => {
```
Ditmar: Test description in Spanish ('muestra estado de error con errorMessage'). Use English. Also the assertion expects aria-invalid attribute; ensure the component explicitly sets aria-invalid on the input when in error state (MUI sets it when error prop is true but if you manipulate inputProps you must ensure it remains).

**Sugerencia:**
```typescript
Translate to English and consider also asserting that helperText is present. Example:

it('shows error state when errorMessage is provided', () => { ... });

And ensure FormInput explicitly sets aria-invalid on the input: inputProps={{ 'aria-label': ..., 'aria-invalid': isError }} or verify MUI behavior in your version.
```

Ditmar (changes requested):
Overall the new FormInput component and accompanying files add useful functionality (variants, sizes, stories, tests and styles). However there are several concrete issues that must be fixed before merging: (1) inconsistency between styled-props (the styled component expects $-prefixed props) and how FormInput passes props (non-$ names) which prevents styles from applying; (2) brittle size/spacing logic in SIZE_CONFIGS and in styles that parses rem strings at runtime and misuses theme.spacing; (3) undefined references (SHAPE_CONFIGS) and missing type imports in types file; (4) usage of default exports (FormInput and Storybook meta) that violate the project's named-export rule; (5) Storybook meta default export conflicts with the project's rule — needs clarification (Storybook requires default export); (6) FormInput docs MDX is malformed (opening code fence not closed / missing newline); (7) tests contain multiple issues: Spanish test descriptions (project requires English), shared mock reuse without reset, untyped test wrapper component (implicit any), and some fragile assertions; (8) theme usage references non-standard theme keys (theme.palette.custom?.focus) which can be undefined and break typing; (9) inputProps merging in the component can be overridden by rest props — needs safe merging. I list each specific line problem below with exact-line suggestions to make the changes concrete.
---

**Conclusión:** Request changes: the PR has useful structure and tests but multiple functional and policy issues must be resolved (styled prop mismatch, brittle size parsing, missing/undefined types, default exports against project rule, malformed docs, test language and mock handling, and usage of non-standard theme keys). After addressing the listed items and running type checks / Storybook / test suite, please submit an updated PR for re-review.

Ditmar (dismissed):

RosaTorrez (changes requested):
#111

RosaTorrez (changes requested):
#111

JoseHuanaco:
Good morning Ing. @Ditmar 
Review please. My PR was approved in QA the monday.

PR #101, opened 2025-10-17
[CDD] Build StatsPanel component

jacob-santos-0:
### 🧾 DESCRIPTION
I implemented the StatsPanel component for the Dashboard.
### COMPONENT OVERVIEW
The component fully meets all the requirements defined in the ticket:
1. ✅ Props: Accepts numerical properties — totalTests, passed, failed, and waiting.
2. 🎨 Design & Styling: Uses MUI icons next to each statistic (work, like, dislike, hourglass) and is styled using the styled() API with theme tokens for colors, background, and spacing.
3. ⚙️ Custom Hook: Includes an internal hook useStatsPanel to handle data management and internal calculations.
4. 🧪 Testing: Added unit tests (StatsPanel.test.tsx) using Vitest to verify that each prop renders correctly.
5. 📘 Storybook: Created Storybook stories (StatsPanel.stories.tsx) showcasing multiple sample data sets.
6. 🧩 Folder Structure: Follows the required component structure:
 StatsPanel/
├── StatsPanel.tsx
├── StatsPanel.types.ts
├── StatsPanel.constants.ts
├── StatsPanel.styles.ts
├── StatsPanel.hook.ts
├── StatsPanel.test.tsx
├── StatsPanel.stories.tsx
└── StatsPanel.docs.mdx (optional)
prueba
prueba

### DIFF
--- a/src/components/StatsPanel/StatsPanel.styles.ts
+++ b/src/components/StatsPanel/StatsPanel.styles.ts
@@ -1,4 +1,5 @@
 import { styled } from '@mui/material/styles';
+
 import type { PaletteColor } from '@mui/material/styles';
 
 export const CardContainer = styled('div')(({ theme }) => ({
--- a/src/components/StatsPanel/StatsPanel.test.tsx
+++ b/src/components/StatsPanel/StatsPanel.test.tsx
@@ -1,8 +1,10 @@
-import { render, screen } from '@testing-library/react';
 import { ThemeProvider } from '@mui/material/styles';
-import theme from '../../style-library/theme/theme';
-import { StatsPanel } from './StatsPanel';
+import { render, screen } from '@testing-library/react';
 import { describe, it, expect } from 'vitest';
+
+import theme from '../../style-library/theme/theme';
+
+import { StatsPanel } from './StatsPanel';
 
 describe('StatsPanel', () => {
   it('renders the provided stats correctly with accessible labels', () => {
--- a/src/style-library/theme/theme.ts
+++ b/src/style-library/theme/theme.ts
@@ -1,8 +1,8 @@
 import Syne700 from '../../assets/fonts/Syne-700.woff2';
 import MontserratSemiBold from '../../components/MCQQuestion/assets/fonts/Montserrat-SemiBold.woff2';
 import SyneRegular from '../../components/MCQQuestion/assets/fonts/Syne-Regular.woff2';
-import SyneSemiBold from '../../components/TestHeader/assets/fonts/Syne-SemiBold.woff2';
 import PoppinsSemiBold from '../../components/StatsPanel/assets/fonts/Poppins-SemiBold.woff2';
+import SyneSemiBold from '../../components/TestHeader/assets/fonts/Syne-SemiBold.woff2';
 import { createCustomTheme } from '../types/theme.helpers';
 
 import type { ThemeConfig } from '../types/theme.types';
@@ -112,7 +112,8 @@ theme.typography = {
     lineHeight: 1.2,
   },
   numbers: {
-    fontFamily: '"Poppins", "Syne", "Montserrat", "Inter", "Roboto", "Helvetica", "Arial", sans-serif',
+    fontFamily:
+      '"Poppins", "Syne", "Montserrat", "Inter", "Roboto", "Helvetica", "Arial", sans-serif',
     fontSize: '1.25rem',
     fontWeight: 600,
     lineHeight: 1.2,


src/components/StatsPanel/StatsPanel.constants.ts
```diff
@@ -0,0 +1,21 @@
+import {
+  WorkOutline,
+  CheckCircleOutline,
+  ThumbDownOffAlt,
+  HourglassEmpty,
+} from '@mui/icons-material';
+
+import type { StatItemData } from './StatsPanel.types';
+
+export const TEST_STATS: Record<string, StatItemData> = {
```
Ditmar: TEST_STATS is typed as Record<string, StatItemData>, which is broad and loses static guarantees about the expected keys (total, passed, failed, waiting). This allows typos in keys to go unnoticed and reduces type-safety when indexing TEST_STATS in the hook.

**Sugerencia:**
```typescript
Narrow the type so the keys are explicit. For example:

export const TEST_STATS: Record<'total' | 'passed' | 'failed' | 'waiting', StatItemData> = {
  ...
};

Or let TypeScript infer the shape by removing the explicit Record type and using a const assertion if appropriate.
```

src/components/StatsPanel/StatsPanel.constants.ts
```diff
@@ -0,0 +1,21 @@
...
+  total: { id: 'total', label: 'No of Tests', value: 32, icon: WorkOutline, color: 'primary' },
```
Ditmar: Copy/label text 'No of Tests' is grammatically awkward and inconsistent with other labels. Use consistent and clear English labels across the UI.

**Sugerencia:**
```typescript
Change the label to 'Total tests' or 'Number of tests', e.g.:

total: { id: 'total', label: 'Total tests', value: 32, icon: WorkOutline, color: 'primary' },
```

src/components/StatsPanel/StatsPanel.constants.ts
```diff
@@ -0,0 +1,21 @@
...
+  passed: { id: 'passed', label: 'Passed', value: 12, icon: CheckCircleOutline, color: 'success' },
+  failed: { id: 'failed', label: 'Failed', value: 19, icon: ThumbDownOffAlt, color: 'error' },
+  waiting: {
+    id: 'waiting',
+    label: 'Waiting for result',
+    value: 1,
+    icon: HourglassEmpty,
+    color: 'grey',
```
Ditmar: The color value 'grey' is used here but this code later indexes theme.palette[color] expecting a PaletteColor with a .main property. MUI's palette does not expose a 'grey' color as a top-level Palette key (theme.palette.grey exists but is not a PaletteColor with .main), so this will likely result in undefined access at runtime or incorrect typing.

**Sugerencia:**
```typescript
Use a palette key that maps to a PaletteColor (e.g. 'text', 'primary', 'secondary', 'error', 'success', 'warning') or change the implementation to accept a function/variant that maps to an explicit theme color. Example options:

- Replace 'grey' with 'text' or 'secondary' (if appropriate).
- Or change StatItemData.color to be a union of allowed keys and/or accept a full color token (like 'grey.500') and update StatIcon to handle it safely.

Example change:
waiting: { id: 'waiting', label: 'Waiting for result', value: 1, icon: HourglassEmpty, color: 'text' },

Or update StatIcon to map 'grey' explicitly to theme.palette.grey[500].
```

src/components/StatsPanel/StatsPanel.hook.ts
```diff
@@ -0,0 +1,20 @@
+import { useMemo } from 'react';
+
+import { TEST_STATS } from './StatsPanel.constants';
+
+import type { StatItemData, StatsPanelProps } from './StatsPanel.types';
+export const useStatsPanel = ({
```
Ditmar: The hook destructures StatsPanelProps directly (no defaults). While this is OK if callers always provide the props, it could be made more defensive. Also the hook relies on TEST_STATS keys always existing; a mismatch between TEST_STATS keys and the destructured names could cause runtime issues.

**Sugerencia:**
```typescript
Consider providing default values or validating inputs in the hook and give TypeScript stronger guarantees on TEST_STATS (see suggestion in constants). Example:

export const useStatsPanel = ({ totalTests = 0, passed = 0, failed = 0, waiting = 0 }: StatsPanelProps): StatItemData[] =>
  useMemo(() => [...], [totalTests, passed, failed, waiting]);

Also ensure TEST_STATS is typed to guarantee 'total', 'passed', 'failed', 'waiting' keys exist.
```

src/components/StatsPanel/StatsPanel.stories.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+const meta: Meta<typeof StatsPanel> = {
+  title: 'Dashboard/StatsPanel',
+  component: StatsPanel,
+  parameters: {
+    layout: 'centered',
+  },
+};
+
+export default meta;
```
Ditmar: This file uses `export default meta;` (Storybook's required default export). Project rules in the repo specify 'Use named exports (NO default exports)'. This is a conflict between the project's rule and Storybook CSF requirements. The rule is strict and this line violates it.

**Sugerencia:**
```typescript
If the project rule is strict, add an exception for Storybook meta default exports in linting or documentation. If no exception will be added, consider documenting the exception in a README for stories. Otherwise, keep the default export because Storybook requires it, but update the project lint/rules to permit this pattern.
```

src/components/StatsPanel/StatsPanel.styles.ts
```diff
@@ -0,0 +1,52 @@
...
+import type { PaletteColor } from '@mui/material/styles';
+
+export const CardContainer = styled('div')(({ theme }) => ({
+  backgroundColor: theme.palette.background.paper,
+  borderRadius: theme.spacing(2.5),
+  padding: theme.spacing(4),
+  display: 'flex',
+  flexDirection: 'column',
+  gap: theme.spacing(3),
+  width: theme.spacing(32.5),
```
Ditmar: CardContainer sets a fixed width using theme.spacing(32.5). This is a very large fixed width expressed via spacing and may cause layout inflexibility and poor responsiveness. Using fractional spacing values also produces fractional pixel values which can cause blurry results in some browsers.

**Sugerencia:**
```typescript
Prefer responsive widths (maxWidth) or use a smaller integer spacing based on the design system. Example:

width: '100%',
maxWidth: theme.spacing(40),

Or choose an integer theme.spacing value: theme.spacing(32) instead of 32.5 and prefer responsive breakpoints if needed.
```

src/components/StatsPanel/StatsPanel.styles.ts
```diff
@@ -0,0 +1,52 @@
...
+  boxShadow: theme.shadows[3],
+}));
+
+export const StatItem = styled('div')(({ theme }) => ({
+  display: 'flex',
+  alignItems: 'center',
+  gap: theme.spacing(1.75),
```
Ditmar: This file uses fractional spacing values (theme.spacing(1.75)). Fractional spacing is allowed but inconsistent with typical design-system spacing scales (usually integer multipliers of theme.spacing). Using fractions can hinder consistency across components.

**Sugerencia:**
```typescript
Prefer the design system spacing scale (integers). Replace fractional values with the closest integer token e.g. theme.spacing(2) or adjust the design tokens if fractional spacing is truly required.
```

src/components/StatsPanel/StatsPanel.styles.ts
```diff
@@ -0,0 +1,52 @@
...
+  height: theme.spacing(6),
+  borderRadius: '50%',
+  display: 'flex',
+  alignItems: 'center',
+  justifyContent: 'center',
+  backgroundColor: theme.palette.action.hover,
+  opacity: 0.9,
+}));
+
+export const StatIcon = styled('div')<{ color: keyof import('@mui/material/styles').Palette }>(
```
Ditmar: The StatIcon styled component declares the prop type as an inline import: { color: keyof import('@mui/material/styles').Palette }. This inline import in a generic is hard to read and could be simplified and made explicit by importing the Palette type at the top.

**Sugerencia:**
```typescript
Import the Palette type at the top for clarity and use it in the generic. Example:

import type { Palette } from '@mui/material/styles';
export const StatIcon = styled('div')<{ color: keyof Palette }>(...)

This improves readability and maintainability.
```

src/components/StatsPanel/StatsPanel.styles.ts
```diff
@@ -0,0 +1,52 @@
...
+  ({ theme, color }) => ({
+    fontSize: theme.typography.h5.fontSize,
+    color: (theme.palette[color] as PaletteColor).main || theme.palette.text.primary,
```
Ditmar: Accessing theme.palette[color] and casting to PaletteColor is unsafe: many keys in Palette are nested objects (e.g. palette.grey) or not PaletteColor with .main. This will lead to runtime errors or undefined. The code then falls back to theme.palette.text.primary but the cast hides the real problem.

**Sugerencia:**
```typescript
Make the color prop a narrower union of valid color keys that map to PaletteColor (e.g. 'primary'|'secondary'|'error'|'warning'|'success') and add a runtime guard. Example:

color: (theme.palette[color] && (theme.palette[color] as PaletteColor).main) || theme.palette.text.primary,

Or accept full CSS color strings and pass them directly without indexing theme.palette. Alternatively, implement a mapping function that converts the incoming color token to a safe theme value (including handling 'grey' specifically via theme.palette.grey[500]).
```

src/components/StatsPanel/StatsPanel.test.tsx
```diff
@@ -0,0 +1,15 @@
+import { render, screen } from '@testing-library/react';
+import { describe, it, expect } from 'vitest';
+
+import { StatsPanel } from './StatsPanel';
+
+describe('StatsPanel', () => {
+  it('renders the provided stats correctly', () => {
+    render(<StatsPanel totalTests={32} passed={12} failed={19} waiting={1} />);
```
Ditmar: The test renders the component with numeric values but there is no check for accessibility attributes (e.g., roles) or labels. The test relies entirely on getByText for values which can be brittle and not representative of accessible usage.

**Sugerencia:**
```typescript
Prefer semantic queries such as getByRole, getByLabelText or add data-testid attributes if necessary. Also assert the presence of labels alongside values. Example change for the test:

// in component: <StatValue aria-label={`stat-value-${stat.id}`}>{stat.value}</StatValue>

// in test:
expect(screen.getByLabelText('stat-value-total')).toHaveTextContent('32');
expect(screen.getByText('Passed')).toBeInTheDocument();

This makes tests more robust and aligned with accessibility.
```

src/components/StatsPanel/StatsPanel.test.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+    expect(screen.getByText('32')).toBeInTheDocument();
```
Ditmar: Using getByText('32') and similar numeric string matches may match other elements unexpectedly (for example if labels or other content contain the same digits). Tests should also assert labels, not only numeric values.

**Sugerencia:**
```typescript
Assert both label and value pairs, or use test ids/aria-labels to target the specific value element. Example:
expect(screen.getByText('Total tests')).toBeInTheDocument();
expect(screen.getByText('32')).toBeInTheDocument();

Or use a specific query like getByLabelText as shown in the previous suggestion.
```

src/components/StatsPanel/StatsPanel.tsx
```diff
@@ -0,0 +1,33 @@
...
+  StatItem,
+  StatIcon,
+  StatText,
+  StatValue,
+  IconWrapper,
+} from './StatsPanel.styles';
+
+import type { StatsPanelProps } from './StatsPanel.types';
+
+export const StatsPanel: React.FC<StatsPanelProps> = ({ totalTests, passed, failed, waiting }) => {
```
Ditmar: The component is typed as React.FC<StatsPanelProps>. Using React.FC introduces the implicit children prop and is discouraged in some TypeScript + React codebases. The project guidelines asked for arrow-function components and named exports — you used an arrow, but React.FC may be unnecessary.

**Sugerencia:**
```typescript
Prefer a plain arrow function with an explicit props annotation to avoid implicit children and clearer typing. Example:

export const StatsPanel = ({ totalTests, passed, failed, waiting }: StatsPanelProps) => {
  ...
};
```

src/components/StatsPanel/StatsPanel.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const stats = useStatsPanel({ totalTests, passed, failed, waiting });
+
+  return (
+    <CardContainer>
+      {stats.map((stat) => (
+        <StatItem key={stat.id}>
+          <IconWrapper>
+            <StatIcon color={stat.color}>{React.createElement(stat.icon)}</StatIcon>
```
Ditmar: Using React.createElement(stat.icon) to render the icon is less idiomatic and loses type inference and the ability to pass props (for example fontSize or aria-hidden). Also stat.icon is typed as ElementType so you can render it as JSX which is clearer.

**Sugerencia:**
```typescript
Use a local variable to render the icon component in JSX and pass props for accessibility and sizing. Example:

const Icon = stat.icon;
<StatIcon color={stat.color}><Icon aria-hidden="true" fontSize="small" /></StatIcon>

Also ensure icons have aria-hidden or an accessible label depending on whether they convey information.
```

src/components/StatsPanel/StatsPanel.tsx
```diff
@@ -0,0 +1,33 @@
...
+          </IconWrapper>
+          <div>
+            <StatValue>{stat.value}</StatValue>
```
Ditmar: StatValue renders a raw number with no formatting or aria attributes. This is fine for simple cases but makes accessibility-based testing and localization harder.

**Sugerencia:**
```typescript
Consider adding an aria-label or data-testid to the value element and format numbers via Intl.NumberFormat if locale formatting is required. Example:

<StatValue aria-label={`stat-value-${stat.id}`}>{stat.value}</StatValue>
```

src/components/StatsPanel/StatsPanel.types.ts
```diff
@@ -0,0 +1,16 @@
...
+  failed: number;
+  waiting: number;
+}
+
+export interface StatItemData {
+  id: string;
+  label: string;
+  value: number;
+  icon: ElementType;
+  color: string;
```
Ditmar: StatItemData.color is typed as string which is too permissive given the component indexes theme.palette[color]. This will allow invalid values (like arbitrary strings) and bypass type-safety.

**Sugerencia:**
```typescript
Narrow the color type to a union of allowed palette keys, for example:

export type StatColor = 'primary' | 'secondary' | 'error' | 'warning' | 'success' | 'info' | 'text';

export interface StatItemData { ... color: StatColor; }

Or import MUI's Palette keys and use keyof Palette if you prefer a broader but typed approach.
```

Ditmar (changes requested):
Overall the new StatsPanel component and its supporting files are well structured and follow many conventions (separate hook, styles, types, tests, and stories). However there are a number of correctness, typing, accessibility, Storybook-convention and robustness issues that need to be addressed before approval. Major points: (1) palette color typing/usage is unsafe — some color values ('grey') will not map to theme.palette[color].main and the StatItemData.color type is too permissive; (2) fragile usage of React.createElement for icons — prefer typed JSX usage to preserve props and accessibility; (3) Storybook default export conflicts with the repo rule requiring named exports only (this is a policy mismatch that must be resolved or documented); (4) tests use non-semantic queries (getByText on numbers) and should use accessible queries or add test ids; (5) styles use some odd spacing/width choices (very large fixed width and fractional spacing values) and an unsafe dynamic palette access in StatIcon; (6) minor copy/text issues (label 'No of Tests') and typing improvements (narrow color type, safer TEST_STATS typing). See below for line-by-line comments and concrete suggestions.
---

**Conclusión:** Requesting changes. The component scaffolding is good but there are multiple type-safety issues (palette color typing and theme indexing), accessibility/testability improvements, Storybook default-export policy mismatch, and a few copy/UX issues that should be fixed before approval. Once colors are typed/mapped safely, icons are rendered via JSX with accessibility attrs, tests use semantic queries, and the Storybook/export policy is reconciled, this PR can be approved.

PR #164, opened 2025-12-05
[CDD] Develop SearchBar component #11

### LINKED ISSUE #11: Develop SearchBar component
### Component purpose
Create a reusable search bar component with a text input and an action button for search.

### Requirements
- The component accepts `placeholder` and `onSearch` props.
- Use a state hook (e.g., `useSearchBar`) to manage input value and handle debouncing.
- Use MUI TextField and IconButton with an SVG search icon, styled via `styled()` or `sx` props.
- Provide a default debounce delay constant; keep constants in `.constants.ts`.
- Expose props typing in `.types.ts`, extending MUI TextField props.
- Provide unit tests verifying callback is invoked after typing and pressing Enter.
- Use `data-testid` attributes for tests.
- Provide Storybook stories demonstrating default state and focused state.
- Document usage and props in `.docs.mdx`.

### File structure
SearchBar/
  ├── SearchBar.tsx
  ├── SearchBar.types.ts
  ├── SearchBar.constants.ts
  ├── SearchBar.hook.ts
  ├── SearchBar.styles.ts
  ├── SearchBar.utils.ts (if needed)
  ├── SearchBar.test.tsx
  ├── SearchBar.stories.tsx
  └── SearchBar.docs.mdx

NatalyGutierrezR27:
SearchBar Component
A reusable and customizable search bar component built with Material-UI. Features debounced input handling, keyboard navigation support, and accessible design patterns.

Overview
The SearchBar provides a user-friendly search interface with:

Debounced search queries to optimize performance
Material-UI integration with consistent styling
Keyboard support (Enter key triggers search)
Accessibility features including ARIA labels
Customizable styling via theme props
Comprehensive testing with user interaction coverage
Basic Usage
import { SearchBar } from './SearchBar';

const MyComponent = () => {
  const handleSearch = (query: string) => {
    console.log('Search query:', query);
  };

  return (
    <SearchBar 
      onSearch={handleSearch}
      placeholder="Search content..."
    />
  );
};
SearchBar component with Material-UI TextField and IconButton, featuring debounced input handling, keyboard support (Enter key), and customizable styling options.
{1C60A746-79C4-4897-A09A-21EF533A84CC}

{45F13183-5127-4150-BD8F-1A15377BB5C6}

https://ditmar.github.io/leetcode-spa/feature-leetcode-11/

Closes #11

### DIFF
--- a/src/components/SearchBar/SearchBar.stories.tsx
+++ b/src/components/SearchBar/SearchBar.stories.tsx
@@ -22,7 +22,7 @@ const meta: Meta<typeof SearchBar> = {
   },
 };
 
-export default meta;
+export default meta; //storybook lo requiere
 
 type Story = StoryObj<typeof SearchBar>;
 
--- a/src/components/SearchBar/SearchBar.test.tsx
+++ b/src/components/SearchBar/SearchBar.test.tsx
@@ -1,5 +1,5 @@
 import { ThemeProvider, createTheme } from '@mui/material/styles';
-import { render, screen } from '@testing-library/react';
+import { render, screen, fireEvent } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import React from 'react';
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
@@ -65,38 +65,35 @@ describe('SearchBar', () => {
     expect(onSearch).toHaveBeenCalledTimes(1);
   });
 
-  it('auto-searches when typing with autoSearch enabled', async () => {
+  it('auto-searches when typing with autoSearch enabled', () => {
     vi.useFakeTimers();
-    const user = userEvent.setup({ delay: null });
     const onSearch = vi.fn();
 
     renderWithTheme(<SearchBar onSearch={onSearch} autoSearch debounceDelay={300} />);
 
     const input = screen.getByRole('textbox');
-    await user.type(input, 'test');
+
+    fireEvent.change(input, { target: { value: 'test' } });
 
     expect(onSearch).not.toHaveBeenCalled();
 
     vi.advanceTimersByTime(300);
     expect(onSearch).toHaveBeenCalledWith('test');
-
-    vi.useRealTimers();
+    expect(onSearch).toHaveBeenCalledTimes(1);
   });
 
-  it('does not auto-search when autoSearch is false', async () => {
+  it('does not auto-search when autoSearch is false', () => {
     vi.useFakeTimers();
-    const user = userEvent.setup({ delay: null });
     const onSearch = vi.fn();
 
     renderWithTheme(<SearchBar onSearch={onSearch} autoSearch={false} debounceDelay={300} />);
 
     const input = screen.getByRole('textbox');
-    await user.type(input, 'test');
+
+    fireEvent.change(input, { target: { value: 'test' } });
 
     vi.advanceTimersByTime(300);
     expect(onSearch).not.toHaveBeenCalled();
-
-    vi.useRealTimers();
   });
 
   it('renders disabled state correctly', () => {


src/components/SearchBar/SearchBar.hook.ts
```diff
@@ -0,0 +1,77 @@
...
+  const {
+    value: controlledValue,
+    defaultValue,
+    onSearch,
+    debounceDelay = DEBOUNCE_DELAY,
+    onChange,
+  } = props;
+
+  const isControlled = controlledValue !== undefined;
+  const [internalValue, setInternalValue] = useState(defaultValue || '');
```
Ditmar: setInternalValue is initialized with a string (defaultValue || ''), so internalValue's inferred type is string. Later you call setInternalValue(controlledValue) where controlledValue can be undefined (props.value is optional). That will cause a TypeScript error or unsafe state assignment. Ensure you never pass undefined to the state setter or type the state to accept undefined explicitly.

**Sugerencia:**
```typescript
Type internal state explicitly and normalize controlledValue when syncing. For example:
const [internalValue, setInternalValue] = useState<string>(defaultValue ?? '');
...
if (isControlled) {
  setInternalValue(controlledValue ?? '');
}
```

src/components/SearchBar/SearchBar.hook.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  // Sync internal state with controlled value
+  useEffect(() => {
+    if (isControlled) {
+      setInternalValue(controlledValue);
+    }
+  }, [controlledValue, isControlled]);
+
+  // Debounce logic
+  useEffect(() => {
```
Ditmar: The debounce effect is present, but the actual onSearch call is commented out. This makes debounceDelay a no-op and is inconsistent with the component docs/stories that advertise debounced input handling. Decide whether the component should auto-trigger onSearch while typing (debounced) or only trigger on Enter/button and update code/docs/tests accordingly.

**Sugerencia:**
```typescript
If you want auto-search on typing, call onSearch with a safe value (see typing fix) inside the debounced handler. If you don't want auto-search, remove debounceDelay and this effect to avoid dead code and surprising API. Alternatively expose an explicit prop (e.g. autoSearch?: boolean) to control behavior.
```

src/components/SearchBar/SearchBar.hook.ts
```diff
@@ -0,0 +1,77 @@
...
+        // onSearch(internalValue); // Uncomment if auto-search is desired
+      }
+    }, debounceDelay);
+
+    return () => {
+      clearTimeout(handler);
+    };
+  }, [internalValue, debounceDelay, onSearch]);
+
+  const handleInputChange = (e: ChangeEvent<HTMLInputElement>) => {
```
Ditmar: Handlers (handleInputChange) are recreated on every render. When this hook is used inside a component, passing new handler references can trigger unnecessary re-renders of child components. Memoize handlers with useCallback and properly declare dependencies.

**Sugerencia:**
```typescript
Wrap the handler in useCallback: const handleInputChange = useCallback((e) => { ... }, [isControlled, onChange]);
```

src/components/SearchBar/SearchBar.hook.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    if (onChange) {
+      onChange(e);
+    }
+  };
+
+  const handleKeyDown = (
+    e: KeyboardEvent<HTMLDivElement | HTMLInputElement | HTMLTextAreaElement>
+  ) => {
+    if (e.key === 'Enter') {
```
Ditmar: When the component is controlled, onSearch currently uses internalValue. If controlledValue is provided by the parent, internalValue may not reflect the latest controlled value (or may be empty). Use a single derived source-of-truth when calling onSearch so controlled usage works reliably.

**Sugerencia:**
```typescript
Create a currentValue const and use it for all search triggers:
const currentValue = isControlled ? (controlledValue ?? '') : internalValue;
... onSearch(currentValue);
This ensures onSearch sends the actual current value in both controlled and uncontrolled modes.
```

src/components/SearchBar/SearchBar.stories.tsx
```diff
@@ -0,0 +1,57 @@
...
+  component: SearchBar,
+  argTypes: {
+    onSearch: { action: 'searched' },
+    onChange: { action: 'changed' },
+    value: { control: 'text' },
+    debounceDelay: { control: 'number' },
+  },
+};
+
+export default meta;
```
Ditmar: This file exports `meta` as the default export (export default meta;). Project rules require named exports only — default exports are disallowed across the codebase.

**Sugerencia:**
```typescript
Export the metadata as a named export and update any Storybook config if required. Example: export const meta = { ... } as Meta<typeof SearchBar>; then update stories to reference `meta` where needed (or rely on CSF's named exports).
```

src/components/SearchBar/SearchBar.styles.ts
```diff
@@ -0,0 +1,128 @@
+import { Box, TextField, IconButton } from '@mui/material';
+import { styled } from '@mui/material/styles';
+
+export const SearchContainer = styled(Box)(({ theme }) => ({
+  position: 'absolute',
```
Ditmar: SearchContainer uses absolute positioning and many hard-coded pixel values and colors (position, width, height, left, top, background, backdropFilter, borderRadius). The project's CDD styling rules require use of theme.spacing(), theme.palette, theme.typography and avoid hardcoded px/hex values so components adapt to theme and are responsive.

**Sugerencia:**
```typescript
Rework layout to use relative, responsive layout (flexbox/grid) inside the parent rather than absolute screen coordinates. Replace hardcoded values with theme tokens, for example:
background: alpha(theme.palette.background.paper, 0.75),
borderRadius: theme.shape.borderRadius,
width: theme.spacing(40) // or responsive variants, etc.
Use theme.breakpoints to adapt sizes instead of fixed left/top coordinates.
```

src/components/SearchBar/SearchBar.styles.ts
```diff
@@ -0,0 +1,128 @@
...
+  height: '67px',
+  left: '929px',
+  top: '91px',
+  background: 'rgba(250, 250, 250, 0.75)', // Fondo con opacidad
+  backdropFilter: 'blur(10px)', // Añadir blur para efecto vidrio
+  borderRadius: '15px',
+  display: 'flex',
+  alignItems: 'center',
+  boxSizing: 'border-box',
+  border: '1px solid rgba(255, 255, 255, 0.2)', // Borde sutil
```
Ditmar: Border is set using a hardcoded rgba string. Use theme.palette/divider or alpha(theme.palette.common.white, x) to keep colors consistent with the theme.

**Sugerencia:**
```typescript
Use the theme token, e.g. border: `1px solid ${alpha(theme.palette.divider, 0.2)}` or borderColor: theme.palette.divider.
```

src/components/SearchBar/SearchBar.styles.ts
```diff
@@ -0,0 +1,128 @@
...
+  [theme.breakpoints.down('sm')]: {
+    width: '342px',
+    height: '67px',
+    left: '24px',
+    top: '231.87px',
+  },
+}));
+
+export const SearchIconButton = styled(IconButton)(({ theme }) => ({
+  position: 'absolute',
```
Ditmar: SearchIconButton uses absolute positioning and very small hard-coded px sizes and offsets. This makes the component fragile and non-themeable. Use theme.spacing and relative placement within a flex container instead.

**Sugerencia:**
```typescript
Place the icon in-flow (no absolute) and use padding/margin derived from theme.spacing. Example: display:flex; align-items:center; paddingLeft: theme.spacing(1); then set the IconButton sizes using theme.spacing values.
```

src/components/SearchBar/SearchBar.styles.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  '& .MuiTouchRipple-root': {
+    display: 'none',
+  },
+
+  '&:hover': {
+    backgroundColor: 'transparent',
+  },
+
+  '& img': {
```
Ditmar: Image icon sizes are hardcoded ('17.81px' etc.). Prefer theme.spacing or MUI icon sizing tokens to ensure visual consistency and scalability.

**Sugerencia:**
```typescript
Use a consistent numeric value (no fractional px), or better use theme.spacing-based sizes: width: theme.spacing(2.5), height: theme.spacing(2.5), or use MUI's SvgIcon + IconButton combination.
```

src/components/SearchBar/SearchBar.styles.ts
```diff
@@ -0,0 +1,128 @@
...
+    top: '24px', // 255.87 - 231.87 = 24px desde el contenedor
+
+    '& img': {
+      width: '18.8px',
+      height: '17.81px',
+    },
+  },
+}));
+
+export const StyledInput = styled(TextField)(({ theme }) => ({
```
Ditmar: StyledInput is absolutely positioned and uses many hard-coded font sizes, line-heights and widths instead of theme.typography and spacing. The CDD rules require typography to come from theme.typography and spacing from theme.spacing.

**Sugerencia:**
```typescript
Let the TextField be layout-managed by its container (no absolute). Use theme.typography for fontFamily/fontSize/lineHeight (e.g. ...theme.typography.body1) and theme.spacing for sizes. Avoid embedding custom fonts inline here; set them at the theme level. If a custom font is required, register it in global styles and reference theme.typography.
```

src/components/SearchBar/SearchBar.types.ts
```diff
@@ -0,0 +1,21 @@
+import type { TextFieldProps } from '@mui/material';
+
+export interface SearchBarProps
```
Ditmar: CDD guidelines require types/props to use a `Props` prefix. The interface is named SearchBarProps rather than starting with 'Props'. Please follow the project's naming convention for prop types.

**Sugerencia:**
```typescript
Rename the interface to follow the prefix convention, for example: export interface PropsSearchBar extends Omit<...> { ... } and update imports.
```

src/components/SearchBar/SearchBar.types.ts
```diff
@@ -0,0 +1,21 @@
...
+  /** Callback function when search is triggered (via button or Enter key) */
+  onSearch?: (value: string) => void;
+  /** Delay in milliseconds for debouncing search calls */
+  debounceDelay?: number;
+  /** Controlled value for the input */
+  value?: string;
+  /** Default value for uncontrolled input */
+  defaultValue?: string;
+  /** Callback when input value changes */
+  onChange?: (event: React.ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => void;
```
Ditmar: There are two props for change events: onChange and inputOnChange. This is confusing (two handlers for the same event). Keep a single, well-named prop for change notifications. The component currently wires both in different ways, which increases cognitive load and chance of bugs.

**Sugerencia:**
```typescript
Drop `inputOnChange` and keep `onChange` (standard naming) — or rename to `onInputChange` if you need to differentiate. Update the component and hook accordingly so a single prop is responsible for notifying consumers.
```

src/components/SearchBar/SearchBar.tsx
```diff
@@ -0,0 +1,73 @@
+import React, { type FC } from 'react';
+
+import searchIcon from './assets/search-icon.svg';
+import { DEFAULT_PLACEHOLDER } from './SearchBar.constants';
+import { useSearchBar } from './SearchBar.hook';
+import { SearchContainer, StyledInput, SearchIconButton } from './SearchBar.styles';
+
+import type { SearchBarProps } from './SearchBar.types';
+
+export const SearchBar: FC<SearchBarProps> = (props) => {
```
Ditmar: The SearchBar component isn't wrapped with React.memo. For a small presentational component that receives handlers from parents, consider wrapping it with React.memo to avoid unnecessary re-renders when parent state changes unrelated to this component.

**Sugerencia:**
```typescript
Create the component as a const and export a memoized version, e.g.:
const SearchBarComponent: FC<PropsSearchBar> = (props) => { ... };
export const SearchBar = React.memo(SearchBarComponent);
```

src/components/SearchBar/SearchBar.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const {
+    placeholder = DEFAULT_PLACEHOLDER,
+    inputOnChange,
```
Ditmar: The component accepts both `inputOnChange` and `onChange` props. This duplicates responsibility and conflicts with the common convention where `onChange` is the single handler for input changes. It also mismatches the types file duplication.

**Sugerencia:**
```typescript
Unify the API: remove `inputOnChange` and use `onChange` consistently, or if you need both give them clearer names and document their responsibilities. Update the hook and tests accordingly.
```

src/components/SearchBar/SearchBar.tsx
```diff
@@ -0,0 +1,73 @@
...
+    if (inputOnChange) {
+      inputOnChange(e);
+    }
+  };
+
+  return (
+    <SearchContainer data-testid="search-bar-container">
+      <SearchIconButton
+        onClick={handleButtonClick}
+        aria-label="search"
```
Ditmar: The button has aria-label="search" while the input uses aria-label "Search content". Accessible names should be consistent and preferably receive their strings from props/i18n so they can be localized. Also consider passing the aria-label as a prop so consumers can customize it.

**Sugerencia:**
```typescript
Use a single label value (from props or i18n) for both button and input where appropriate, e.g. aria-label={ariaLabel ?? t('search')}. Ensure casing/phrasing matches to prevent confusing screen reader output.
```

src/components/SearchBar/SearchBar.tsx
```diff
@@ -0,0 +1,73 @@
...
+        data-testid="search-button"
+        disabled={disabled}
+      >
+        <img src={searchIcon} alt="Search" data-testid="search-icon" />
```
Ditmar: The  inside the IconButton has alt="Search". Because the button already has an accessible name (aria-label), the image should be decorative to avoid duplicate announcements. Either use an empty alt attribute (alt="") or mark the image aria-hidden.

**Sugerencia:**
```typescript
Make the image decorative: <img src={searchIcon} alt="" aria-hidden="true" /> or use MUI SvgIcon so the icon is accessible via the button label only.
```

src/components/SearchBar/SearchBar.tsx
```diff
@@ -0,0 +1,73 @@
...
+      </SearchIconButton>
+      <StyledInput
+        value={inputValue}
+        onChange={combinedHandleChange}
+        placeholder={placeholder}
+        variant="standard"
+        inputProps={{
+          'aria-label': 'Search content',
```
Ditmar: The input's aria-label and other strings are hardcoded in the component. Project i18n rules require strings to be localizable / not hardcoded. Provide these strings via props or a translation function.

**Sugerencia:**
```typescript
Accept an `ariaLabel` prop and/or use an i18n function: inputProps={{ 'aria-label': ariaLabel ?? t('searchContent') }}. Also localize placeholder if needed.
```

src/components/SearchBar/SearchBar.constants.ts
```diff
@@ -0,0 +1,2 @@
+export const DEBOUNCE_DELAY = 300;
+export const DEFAULT_PLACEHOLDER = 'Search';
```
Ditmar: DEFAULT_PLACEHOLDER is a hardcoded English string. Per project i18n/L10n rules, UI text should be localizable or provided by consumers via props.

**Sugerencia:**
```typescript
Either document this as a fallback and accept placeholder via props (already present), or make this constant come from a locale map or translation helper. At minimum, mark it as a fallback and avoid using it directly in consumer code if localization is required.
```

src/components/SearchBar/SearchBar.docs.mdx
```diff
@@ -0,0 +1,36 @@
...
+    console.log('Search query:', query);
+  };
+
+  return (
+    <SearchBar 
+      onSearch={handleSearch}
+      placeholder="Search content..."
+    />
+  );
+};
```
Ditmar: File ends with "No newline at end of file" in the patch. Add a trailing newline to comply with standard POSIX/Unix text file conventions and avoid noise in diffs.

**Sugerencia:**
```typescript
Add a newline at EOF.
```

Ditmar (changes requested):
styles heavily rely on absolute pixel values, hardcoded colors/fonts and positioning which violates the project's CDD styling requirements (must use theme.palette, spacing, typography and responsive layout instead of magic px values). (2) The hook has a potential TypeScript type bug when syncing controlled value into internal state (setting state with possibly undefined), and the debounce behavior is inconsistent: there's a debounce prop but the effect intentionally comments-out the auto-search call (docs and constants promise debounced behaviour). Decide whether the component should auto-search while typing or only search on Enter/button — implement one behaviour and keep tests/docs aligned. (3) API/naming inconsistencies: the props include both onChange and inputOnChange which is confusing; prefer a single clearly named prop. (4) Storybook file uses a default export (meta) — project rules require named exports. (5) Accessibility and a11y improvements: the icon  inside the button shouldn't create duplicate accessible names (use alt="" for decorative icons or aria-hidden), and aria-labels should be consistent and localizable. (6) Performance: handlers are not memoized; recommend useCallback/useMemo or React.memo the component. I list specific, line-precise comments below with suggestions and examples for fixes.
---

**Conclusión:** This PR introduces the SearchBar and related assets/tests/stories, but there are important issues to address before approval: fix the hook's controlled-state typing, decide and implement consistent debounce behavior, refactor styles to use theme tokens and responsive layout (avoid absolute px magic), unify prop naming (remove inputOnChange duplication), address Storybook default export, and improve a11y (decorative icon alt/aria and consistent labels). After these changes, re-run lint/build/tests and resubmit for final review.

PR #165, opened 2025-12-06
Build SettingsPage

Adrex-2003:
image
image

**What the component does:**
- Allows switching the **theme** between light and dark mode.
- Enables or disables **notifications**.
- Selects the **interface language**.
- Adjusts **privacy options**: public profile, share activity, and save history.

**How it works:**
- Uses MUI for the UI (Switches, Select, Accordion for mobile layout).
- Provides two layouts: **desktop** and **mobile**, adapting to screen size.
- Each option has `aria-label`, `id`, and `data-testid` attributes to ensure accessibility and reliable testing.
- Internal logic is handled via the `useSettingsPage` hook.

**Important notes:**
- Only **two stories** are included: theme change and language change, as these are the main, manageable functionalities for demonstration.
- `NavigationSidebar` is not used because this component does not yet exist in the project.
- Accessibility improvements and strict typing were added, especially for the privacy switches.

With this PR, the component is fully documented, testable, and ready for integration, following best practices for accessibility and visual consistency.

### DIFF
--- a/src/pages/SettingsPage/SettingsPage.docs.mdx
+++ b/src/pages/SettingsPage/SettingsPage.docs.mdx
@@ -1,7 +1,7 @@
 import { Meta, Canvas } from '@storybook/blocks';
 import * as SettingsStories from './SettingsPage.stories';
 
-<Meta title="Pages/Settings Page" />
+<Meta title="Pages/SettingsPage" />
 
 # Settings Page
 
--- a/src/pages/SettingsPage/SettingsPage.stories.tsx
+++ b/src/pages/SettingsPage/SettingsPage.stories.tsx
@@ -6,7 +6,7 @@ import type { SettingsPageProps } from './SettingsPage.types';
 import type { Meta, StoryObj } from '@storybook/react';
 
 const meta: Meta<typeof SettingsPage> = {
-  title: 'Pages/Settings',
+  title: 'Pages/SettingsPage',
   component: SettingsPage,
   parameters: {
     layout: 'fullscreen',


src/pages/SettingsPage/SettingsPage.hook.ts
```diff
@@ -0,0 +1,125 @@
...
+} satisfies UserPreferences;
+
+const migratePreferences = (raw: unknown): UserPreferences => {
+  if (!raw || typeof raw !== 'object') return DEFAULT_PREFERENCES;
+  return { ...DEFAULT_PREFERENCES, ...(raw as Partial<UserPreferences>) };
+};
+
+const loadInitialPreferences = (initialFromProps?: Partial<UserPreferences>): UserPreferences => {
+  try {
+    const item = localStorage.getItem(SETTINGS_LOCAL_STORAGE_KEY);
```
Ditmar: This reads localStorage during the function call. loadInitialPreferences is used inside a useState initializer, which executes during rendering — this will also run on the server during SSR and will throw (localStorage is undefined).

**Sugerencia:**
```typescript
Guard access to localStorage with a runtime check (typeof window !== 'undefined') or move the client-only loading into a useEffect that sets state after mount. A safe approach: initialize state with DEFAULT_PREFERENCES (merged with initialPreferences) and in a useEffect read localStorage when window is available, then update state if needed.
```

src/pages/SettingsPage/SettingsPage.hook.ts
```diff
@@ -0,0 +1,125 @@
...
+    console.warn('[useSettingsPage] Error reading localStorage:', error);
+  }
+
+  return initialFromProps ? { ...DEFAULT_PREFERENCES, ...initialFromProps } : DEFAULT_PREFERENCES;
+};
+
+export const useSettingsPage = (props?: SettingsPageProps): UseSettingsPageReturn => {
+  const { initialPreferences } = props ?? {};
+
+  const [preferences, setPreferences] = useState<UserPreferences>(() =>
```
Ditmar: The state initializer calls loadInitialPreferences which (currently) reads localStorage synchronously — see the SSR concern above. Initializing state with a function that accesses browser-only globals can cause runtime exceptions in non-browser environments.

**Sugerencia:**
```typescript
Use a safe initial value (DEFAULT_PREFERENCES merged with props.initialPreferences) and perform localStorage hydration in a client-only effect. For example, use useEffect(() => { /* read localStorage and setPreferences */ }, []);
```

src/pages/SettingsPage/SettingsPage.tsx
```diff
@@ -0,0 +1,263 @@
...
+                  key === 'publicProfile'
+                    ? 'privacyPublicProfile'
+                    : key === 'shareActivity'
+                      ? 'privacyShareActivity'
+                      : 'privacySaveHistory'
+                ]
+              }
+            </Typography>
+            <OptionDescription>
+              {key === 'publicProfile' && 'Tu perfil será visible para otros usuarios'}
```
Ditmar: Privacy descriptions are hard-coded Spanish strings in the JSX rather than being sourced from the SETTINGS_LABELS i18n object. This bypasses the existing i18n system and will break localization.

**Sugerencia:**
```typescript
Move these descriptions into your SETTINGS_LABELS translations (e.g., privacyPublicProfileDescription, privacyShareActivityDescription, privacySaveHistoryDescription) and reference t.someKey here instead of hardcoded strings.
```

src/pages/SettingsPage/SettingsPage.tsx
```diff
@@ -0,0 +1,263 @@
...
+        </AccordionSummary>
+        <AccordionDetails>
+          <FormControl fullWidth>
+            <InputLabel id="mobile-language-select-label">{t.language}</InputLabel>
+            <Select
+              labelId="mobile-language-select-label"
+              value={preferences.language}
+              label={t.language}
+              // eslint-disable-next-line @typescript-eslint/no-explicit-any
+              onChange={(e) => setLanguage(e.target.value as any)}
```
Ditmar: Same issue as above: using an any cast for the mobile Select onChange handler.

**Sugerencia:**
```typescript
Use SelectChangeEvent (typed) and narrow e.target.value to Locale instead of casting to any.
```

PR #211, opened 2026-04-02
implement InputOTP

### LINKED ISSUE #193: InputOTP — MUI CDD Implementation
## Summary
Implement the `InputOTP` component following CDD principles using Material UI. The component collects a one-time password or verification code via individual digit boxes and must be easy to use on mobile (especially with the numeric soft keyboard).

## Type
Task

## Priority
Medium

## Labels
- CDD
- MUI
- Component
- Frontend
- Responsive

---

## Figma Reference
[View Figma Design](https://www.figma.com/design/1EYOnbPan3uCJFjlUbeGXc/Untitled?node-id=0-1&t=sSJuoUv8FEzXRTvJ-1)

---

## Description
Create an `InputOTP` component composed of `n` individual MUI `TextField` or `Input` boxes (one per digit). The component handles auto-focus advancement, backspace regression, paste-to-fill, and mobile numpad opening. Each digit box must meet the touch-friendly size requirement on mobile.

---

## Mandatory File Structure

```
InputOTP/
├── InputOTP.tsx              # Main component logic & JSX
├── InputOTP.types.ts         # Props interfaces (length, value, onChange, etc.)
├── InputOTP.constants.ts     # Default length, allowed character pattern (optional)
├── InputOTP.hook.ts          # Focus management, paste handling, value state hook
├── InputOTP.styles.ts        # MUI styled() / sx styles — no external CSS, no hardcoded px
├── InputOTP.utils.ts         # Paste splitting, numeric validation helpers (optional)
├── InputOTP.test.tsx         # Unit tests — Jest / Testing Library
├── InputOTP.stories.tsx      # Storybook stories in CSF format
└── InputOTP.docs.mdx         # MDX documentation for Storybook
```

---

## Acceptance Criteria
- [ ] Component implemented using Material UI input primitives
- [ ] Configurable `length` prop (default 6)
- [ ] Auto-advances focus to next box on character entry
- [ ] Backspace moves focus to previous box and clears it
- [ ] Paste event fills all boxes from the pasted value
- [ ] `inputMode="numeric"` and `pattern="[0-9]*"` for mobile numpad
- [ ] `disabled` and `error` states with matching visual feedback
- [ ] Fully responsive — boxes scale via flex/percentage, no overflow at 320px
- [ ] Each box min 44×44px on mobile
- [ ] Tested at 320px, 375px, 768px, 1024px
- [ ] TypeScript types defined in `InputOTP.types.ts`
- [ ] Styles in `InputOTP.styles.ts` using `styled()` or `sx` — no external CSS, no hardcoded `px`
- [ ] Unit tests covering input, auto-advance, backspace, paste, and error state
- [ ] Storybook stories in CSF format
- [ ] MDX documentation written in English
- [ ] Passes axe / ARIA accessibility checks (`aria-label` per box, `role="group"`)
- [ ] Figma design fidelity verified

---

## Technical Notes
- Render `n` controlled `<input>` elements wrapped in MUI `OutlinedInput` or `Box`
- Prevent non-numeric characters via `onKeyDown` guard
- Use MUI `theme.spacing` for box size — no hardcoded `px`
- All comments and documentation must be written in **English**

rodrigopaco:


src/component-catalog/InputOTP/InputOTP.hook.ts
```diff
@@ -0,0 +1,81 @@
+import { useState, useRef } from 'react';
+
+import { OTP_SINGLE_DIGIT_REGEX, OTP_SANITIZE_REGEX } from './InputOTP.constants';
+
+export function useInputOTP(length: number, disabled?: boolean) {
+  const [values, setValues] = useState<string[]>(Array(length).fill(''));
+
+  const inputRefs = useRef<(HTMLInputElement | null)[]>([]);
```
alecuestasvp: Severity: medium
It's best to initialize it with the exact size if the length is static; it could be something like this.
`const inputRefs = useRef<(HTMLInputElement | null)[]>(new Array(length).fill(null));`

src/component-catalog/InputOTP/InputOTP.hook.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    e.preventDefault();
+
+    const pasteData = e.clipboardData.getData('text');
+    const sanitized = pasteData.replace(OTP_SANITIZE_REGEX, '');
+
+    if (!sanitized) return;
+
+    const newValues = [...values];
+
```
alecuestasvp, line 68: Severity: medium
If you paste into input 3, it still starts from 0. This could be improved by using the current index, which would have to be passed as a parameter.

src/component-catalog/InputOTP/InputOTP.stories.tsx
```diff
@@ -0,0 +1,46 @@
+import { InputOTP } from './InputOTP';
+
+import type { Meta, StoryObj } from '@storybook/react';
+
+const meta: Meta<typeof InputOTP> = {
+  title: 'component-catalog/InputOTP',
+  component: InputOTP,
+  decorators: [
```
alecuestasvp: Severity: low
Perhaps this segment could be removed as it may be unnecessary and could break the layout

src/component-catalog/InputOTP/InputOTP.styles.ts
```diff
@@ -0,0 +1,82 @@
+import { styled, Box, TextField, Typography } from '@mui/material';
+
+export const Label = styled(Typography)(() => ({
+  fontFamily: 'Inter, sans-serif',
```
alecuestasvp: Severity: Medium
It breaks the consistency of the design system; if the theme changes, this component is not updated.
`fontFamily: theme.typography.fontFamily,`

src/component-catalog/InputOTP/InputOTP.styles.ts
```diff
@@ -0,0 +1,82 @@
...
+  fontWeight: 500,
+  fontSize: '0.875rem',
+  lineHeight: '0.875rem',
+  color: 'var(--foreground)',
```
alecuestasvp, line 8: Severity: High
It can completely break the UI; this could be improved.
`color: 'var(--foreground, #000)',`

src/component-catalog/InputOTP/InputOTP.styles.ts
```diff
@@ -0,0 +1,82 @@
...
+interface StyledProps {
+  isFirst?: boolean;
+  isLast?: boolean;
+}
+
+export const StyledTextField = styled(TextField, {
+  shouldForwardProp: (prop) => prop !== 'isFirst' && prop !== 'isLast',
+})<StyledProps>(({ theme, isFirst, isLast }) => ({
+  flex: 1,
+  minWidth: theme.spacing(5),
```
alecuestasvp: Severity: Medium
You're limiting the size too much; it doesn't fit well with: small phones, user zoom, accessibility
`minWidth: 40,
maxWidth: 56,`

src/component-catalog/InputOTP/InputOTP.tsx
```diff
@@ -0,0 +1,52 @@
...
+  error = false,
+}: InputOTPProps) {
+  const { values, inputRefs, handleChange, handleKeyDown, handlePaste } = useInputOTP(
+    length,
+    disabled
+  );
+
+  return (
+    <Container role="group" aria-label="One-time password input">
+      <Label>Enter OTP Code</Label>
```
alecuestasvp: Severity: Medium
The code is not reusable, meaning you cannot change the text.
`<Label>{label || 'Enter OTP Code'}</Label>`

src/component-catalog/InputOTP/InputOTP.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { Container, StyledTextField, Label } from './InputOTP.styles';
+
+import type { InputOTPProps } from './InputOTP.types';
+
+export function InputOTP({
+  length = DEFAULT_OTP_LENGTH,
+  disabled = false,
+  error = false,
+}: InputOTPProps) {
+  const { values, inputRefs, handleChange, handleKeyDown, handlePaste } = useInputOTP(
```
alecuestasvp, line 17: Severity: High
You cannot react when the OTP is complete (automatic login, validation)
`useInputOTP(length, disabled, onComplete)`

src/component-catalog/InputOTP/InputOTP.stories.tsx
rodrigopaco: Thanks for the suggestion!

In this case, I kept the decorator intentionally since it was provided as part of the course guidelines to improve component visibility during development in Storybook.

I understand that in a production-ready component library this could be removed or replaced with Storybook layout parameters, but for this context it helps with consistent visualization.

If needed, I can refactor it later to use Storybook's `layout: 'centered'` parameter instead.

src/component-catalog/InputOTP/InputOTP.styles.ts
rodrigopaco: The CSS variable is part of a global design token system (defined in :root and .dark),
so it is always available and safe to use.

src/component-catalog/InputOTP/InputOTP.constants.ts
```diff
@@ -0,0 +1,4 @@
+export const DEFAULT_OTP_LENGTH = 6;
+export const OTP_SINGLE_DIGIT_REGEX = /^[0-9]?$/;
```
Ditmar, line 3: medium: OTP_SINGLE_DIGIT_REGEX is defined as /^[0-9]?$/, which allows an empty string. The hook's handleChange uses this regex to validate input, meaning empty values pass. This may be intentional to allow deletion, but it's not obvious.
Suggestion: If the goal is to only accept a single digit or empty (for deletions) keep this but document it; otherwise change to /^[0-9]$/ to only accept a single digit.

**Suggestion:**
```typescript
If intended to accept only digits (non-empty), use /^[0-9]$/. If accepting empty values on change is intentional for deletion flows, add a short comment in the file to explain that behavior.
```

src/component-catalog/InputOTP/InputOTP.docs.mdx
```diff
@@ -0,0 +1,97 @@
...
+- Paste full OTP support
+- Numeric-only input
+- Disabled and error states
+- Fully responsive layout
+
+---
+
+## Basic Usage
+
+  <Story of={InputOTPStories.Primary} />
```
Ditmar, line 31: high: The MDX uses <Story of={InputOTPStories.Primary} /> but InputOTPStories is not defined or imported here. This MDX is not a valid Storybook docs file in CSF/MDX format and will fail to render in Storybook.
Suggestion: Use correct MDX CSF patterns: import Meta and Story, import the component and the story exports (or embed a Canvas/Story block). Example:
import { Meta, Story } from '@storybook/addon-docs';
import { Primary } from './InputOTP.stories';
<Story name="Primary">{Primary}</Story>

**Suggestion:**
```typescript
Replace the invalid <Story .../> usage with a valid MDX story block and import the story or component explicitly. Example:
import { Meta, Story } from '@storybook/react';
import { Primary } from './InputOTP.stories';
<Story of={Primary} />
```

src/component-catalog/InputOTP/InputOTP.docs.mdx
```diff
@@ -0,0 +1,97 @@
...
+| value    | string                   | —      | Controlled value (optional) |
+| onChange | (value: string) => void  | —      | Callback when value changes |
+
+---
+
+## Examples
+
+### Default (6 digits)
+
+<InputOTP />
```
Ditmar, line 53: medium: The examples in this MDX (e.g. <InputOTP /> and other JSX snippets) are used without importing the InputOTP component or providing a proper MDX/CSF context. This will cause the MDX to fail to render and provides a poor developer experience.
Suggestion: Add proper imports at the top of the MDX: import { InputOTP } from './InputOTP'; import { Meta } from '@storybook/react'; and a default export Meta as needed.

**Suggestion:**
```typescript
Add imports and Meta block to the MDX file and/or use <Preview> / <Canvas> blocks around the example components.
```

src/component-catalog/InputOTP/InputOTP.docs.mdx
```diff
@@ -0,0 +1,97 @@
...
+
+- Only numeric input is allowed
+- Pasted values are sanitized (non-numeric removed)
+- Focus automatically moves between inputs
+
+---
+
+## Design Considerations
+
+- Built with Material UI
```
Ditmar, line 123: medium: The docs claim 'Uses theme.spacing (no hardcoded px)' but the styles file contains hardcoded font sizes (e.g. '0.875rem'). This is an inconsistency between docs and implementation.
Suggestion: Update docs to reflect reality or change the styles to use theme tokens (theme.typography or theme.spacing) consistently, removing hardcoded rem values.

**Suggestion:**
```typescript
Either change styles to use theme.typography variants (e.g. theme.typography.body2.fontSize) or update docs to note places where rem values are used and why.
```

src/component-catalog/InputOTP/InputOTP.hook.ts
```diff
@@ -0,0 +1,91 @@
+import { useState, useRef } from 'react';
```
Ditmar: high: The file references React types (React.ChangeEvent, React.KeyboardEvent, React.ClipboardEvent) but does not import React or import those types explicitly. TypeScript will likely error unless global React types are available. Import explicit types instead.
Suggestion: Import the event types from React using `import type { ChangeEvent, KeyboardEvent, ClipboardEvent } from 'react';` and update signatures to use these types, or add `import React from 'react';` if you prefer.

**Suggestion:**
```typescript
Add at the top: import type { ChangeEvent, KeyboardEvent, ClipboardEvent } from 'react'; and replace occurrences (e.g. e: React.ChangeEvent<...>) with e: ChangeEvent<...>.
```

src/component-catalog/InputOTP/InputOTP.hook.ts
```diff
@@ -0,0 +1,91 @@
...
+
+import { OTP_SINGLE_DIGIT_REGEX, OTP_SANITIZE_REGEX } from './InputOTP.constants';
+
+export function useInputOTP(
+  length: number,
+  disabled?: boolean,
+  onComplete?: (value: string) => void
+) {
+  const [values, setValues] = useState<string[]>(Array(length).fill(''));
```
Ditmar, line 12: high: The hook initializes state with Array(length).fill('') but does not handle changes to the incoming length parameter. If length changes after mount the values state will be out of sync with the requested length and refs will be wrong.
Suggestion: Add a useEffect to reset the values and inputRefs when length changes (and ensure the component supports dynamic length changes), and type the returned structure explicitly.

**Suggestion:**
```typescript
Use useEffect to run when 'length' changes: setValues(Array(length).fill('')); update inputRefs.current = new Array(length).fill(null);
```

src/component-catalog/InputOTP/InputOTP.hook.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  const inputRefs = useRef<(HTMLInputElement | null)[]>(new Array(length).fill(null));
```
Ditmar: high: inputRefs initialization uses new Array(length).fill(null) only on first render; when length changes the refs array will not be updated. Also the typed ref type is fine but ensure you re-create/resize it when length changes.
Suggestion: In a useEffect triggered by length changes, reset inputRefs.current to a proper length and preserve existing references if necessary.

**Suggestion:**
```typescript
After adding the useEffect for length changes, set inputRefs.current = Array(length).fill(null) or use a stable ref array resize helper.
```

src/component-catalog/InputOTP/InputOTP.hook.ts
```diff
@@ -0,0 +1,91 @@
...
+    e: React.ChangeEvent<HTMLInputElement | HTMLTextAreaElement>,
+    index: number
+  ) => {
+    if (disabled) return;
+
+    const val = e.target.value;
+
+    if (!OTP_SINGLE_DIGIT_REGEX.test(val)) return;
+
+    const newValues = [...values];
```
Ditmar: high: The hook uses the spread operator ([...values]) to clone arrays in multiple places (lines 24, 41, 48, 66). The project 'GENERAL FORBIDDEN PATTERNS' forbids use of the spread operator.
Suggestion: Replace spread usage with Array.prototype.slice() or Array.from() to clone arrays (e.g. values.slice() or Array.from(values)).

**Suggestion:**
```typescript
Replace `const newValues = [...values];` with `const newValues = values.slice();` or `Array.from(values);` across the file.
```

src/component-catalog/InputOTP/InputOTP.hook.ts
```diff
@@ -0,0 +1,91 @@
...
+    if (!newValues.includes('')) {
+      onComplete?.(newValues.join(''));
+    }
+
+    if (val && index < length - 1) {
+      inputRefs.current[index + 1]?.focus();
+    }
+  };
+
+  const handleKeyDown = (e: React.KeyboardEvent, index: number) => {
```
Ditmar: medium: The handleKeyDown signature uses a general React.KeyboardEvent but doesn't type the element generic (e.g. HTMLInputElement). Be explicit to improve type-safety and clarity.
Suggestion: Use KeyboardEvent<HTMLInputElement> instead of the generic KeyboardEvent.

**Suggestion:**
```typescript
Change signature to: const handleKeyDown = (e: KeyboardEvent<HTMLInputElement>, index: number) => { ... } after importing KeyboardEvent from 'react'.
```

src/component-catalog/InputOTP/InputOTP.hook.ts
```diff
@@ -0,0 +1,91 @@
...
+        inputRefs.current[index - 1]?.focus();
+      }
+    }
+  };
+
+  const handlePaste = (e: React.ClipboardEvent, index: number) => {
+    if (disabled) return;
+
+    e.preventDefault();
+    const pasteData = e.clipboardData.getData('text');
```
Ditmar, line 62: low: handlePaste uses e.clipboardData.getData('text') which is fine, but the hook prevents default always. Consider whether you should only prevent default for certain paste contents or when handling.
Suggestion: Keep preventing default but add a short comment to explain rationale so future maintainers understand why default paste is prevented.

**Suggestion:**
```typescript
Add a comment before e.preventDefault(): // Prevent default paste so we can sanitize and distribute pasted digits across inputs
```

src/component-catalog/InputOTP/InputOTP.stories.tsx
```diff
@@ -0,0 +1,46 @@
+import { InputOTP } from './InputOTP';
```
Ditmar, line 3: medium: Import ordering is inverted: local component import appears before external Storybook types import. The team's import ordering convention requires external imports first, then internal.
Suggestion: Move `import type { Meta, StoryObj } from '@storybook/react';` to be the first import (before local imports) or group external imports first.

**Suggestion:**
```typescript
Reorder imports:
import type { Meta, StoryObj } from '@storybook/react';
import { InputOTP } from './InputOTP';
```

src/component-catalog/InputOTP/InputOTP.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+import type { Meta, StoryObj } from '@storybook/react';
+
+const meta: Meta<typeof InputOTP> = {
+  title: 'component-catalog/InputOTP',
+  component: InputOTP,
+  decorators: [
+    (Story) => (
```
Ditmar: medium: The decorator uses an inline anonymous function rendering JSX and an inline style object. The team forbids inline anonymous functions inside JSX when avoidable and discourages inline styles/hardcoded layout values in stories.
Suggestion: Move the decorator to a named function exported in the file (or a shared decorators file) and use a CSS class or theme-aware wrapper component instead of inline styles. Avoid '100vh' hardcoded value; prefer a token or leave layout to Storybook canvas.

**Suggestion:**
```typescript
Create a named decorator: const CenterDecorator = (Story) => <CenterWrapper><Story/></CenterWrapper>; and export/use that. Replace inline style with a styled Box or CSS class making use of theme spacing.
```

src/component-catalog/InputOTP/InputOTP.styles.ts
```diff
@@ -0,0 +1,82 @@
+import { styled, Box, TextField, Typography } from '@mui/material';
+
+export const Label = styled(Typography)(({ theme }) => ({
+  fontFamily: theme.typography.fontFamily,
+  fontWeight: 500,
+  fontSize: '0.875rem',
```
Ditmar: medium: Label fontSize and lineHeight are hardcoded as '0.875rem'. The project's styling conventions prefer using theme tokens (theme.typography variants) rather than hardcoded rem values.
Suggestion: Use theme typography values, e.g. fontSize: theme.typography.body2.fontSize, lineHeight: theme.typography.body2.lineHeight, or use a token exported from the design system.

**Suggestion:**
```typescript
Replace hardcoded values with theme properties: fontSize: theme.typography.body2.fontSize, lineHeight: theme.typography.body2.lineHeight.
```

src/component-catalog/InputOTP/InputOTP.styles.ts
```diff
@@ -0,0 +1,82 @@
...
+  color: 'var(--foreground)',
+}));
+
+export const Container = styled(Box)(({ theme }) => ({
+  display: 'flex',
+  flexDirection: 'column',
+  alignItems: 'flex-start',
+  justifyContent: 'center',
+  gap: theme.spacing(1),
+  width: 'fit-content',
```
Ditmar: low: Container uses 'width: "fit-content"' and 'margin: "0 auto"' which are fine but are hardcoded string values. Consider whether these layout choices belong in a presentational component vs. story/layout wrapper.
Suggestion: If this component can be reused across layouts, consider removing the centering/margin and let callers decide layout, or make them optional via props.

**Suggestion:**
```typescript
Consider exposing a prop or omitting the margin so parent components can control centering.
```

src/component-catalog/InputOTP/InputOTP.styles.ts
```diff
@@ -0,0 +1,82 @@
...
+interface StyledProps {
+  isFirst?: boolean;
+  isLast?: boolean;
+}
+
+export const StyledTextField = styled(TextField, {
+  shouldForwardProp: (prop) => prop !== 'isFirst' && prop !== 'isLast',
+})<StyledProps>(({ theme, isFirst, isLast }) => ({
+  flex: 1,
+  minWidth: theme.spacing(5.5),
```
Ditmar: medium: Using theme.spacing with decimal (5.5) is unusual. MUI spacing supports numeric input but decimal usage could be confusing. Confirm that the intended sizes map correctly to design tokens and are consistent.
Suggestion: Prefer whole-number spacing multiples or use pixel/typography tokens for precise control.

**Suggestion:**
```typescript
If 5.5 * spacing unit is intentional, consider using a constant (e.g. theme.spacing(6) with a different minWidth) or a design token to improve readability.
```

src/component-catalog/InputOTP/InputOTP.styles.ts
```diff
@@ -0,0 +1,82 @@
...
+  maxWidth: theme.spacing(6),
+
+  '& .MuiOutlinedInput-root': {
```
Ditmar: high: The styles rely on MUI internal class names such as '.MuiOutlinedInput-root', '.Mui-focused', and '.Mui-error'. The team's conventions explicitly discourage relying on MUI internals because they can change and break styles.
Suggestion: Use the TextField's slotProps, componentsProps, or the styled API in a way that targets component slots via the component's prop overrides (e.g., using `components`/`componentsProps` or the `sx` prop) rather than hardcoded class names. Alternatively, style the input element directly via the styled wrapper's slot props.

**Suggestion:**
```typescript
Refactor to use the TextField's slotProps (e.g. <TextField InputProps={{ classes: { root: ... }}}/>) or use `styled('input')` for the inner input rather than targeting MUI internal class names.
```

src/component-catalog/InputOTP/InputOTP.styles.ts
```diff
@@ -0,0 +1,82 @@
...
+    '&.Mui-error fieldset': {
+      borderColor: 'var(--destructive)',
+    },
+  },
+
+  '& input': {
+    textAlign: 'center',
+    padding: 0,
+    height: '100%',
+    fontSize: theme.typography.h6.fontSize,
```
Ditmar: medium: theme.typography.h6.fontSize is accessed directly. If the theme does not define h6 or its fontSize, this may be undefined. Prefer using a defined typography variant or fallback.
Suggestion: Use theme.typography.h6?.fontSize || theme.typography.fontSize or a named variant to ensure presence.

**Suggestion:**
```typescript
Change to fontSize: theme.typography.h6?.fontSize ?? theme.typography.body1.fontSize to provide a safe fallback.
```

src/component-catalog/InputOTP/InputOTP.test.tsx
```diff
@@ -0,0 +1,82 @@
+import { render, screen, fireEvent } from '@testing-library/react';
```
Ditmar, line 1: medium: Tests import from @testing-library/react which is fine, but the project devDependencies appear to use pinned versions inconsistently (package.json changes). The test runner confusion (vitest vs jest) may make these tests fail in CI. Also, tests do not assert the onComplete callback being called when OTP is filled.
Suggestion: Ensure test runner alignment in package.json (use vitest or jest consistently) and add tests to assert onComplete behavior for the full-OTP paste/typing flow.

**Suggestion:**
```typescript
Add a test that supplies a jest/vitest mock for onComplete and validate it is invoked with the expected OTP after inputs are filled or when paste completes.
```

src/component-catalog/InputOTP/InputOTP.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+    render(<InputOTP length={6} />);
+    const inputs = screen.getAllByRole('textbox');
+    expect(inputs).toHaveLength(6);
+  });
+
+  test('allows typing a digit and moves focus forward', () => {
+    render(<InputOTP length={4} />);
+    const inputs = screen.getAllByRole('textbox');
+
+    fireEvent.change(inputs[0], { target: { value: '1' } });
```
Ditmar: low: Tests use fireEvent.change and fireEvent.keyDown. Prefer user-event for more realistic user interactions (and for better alignment with best practices). Using userEvent also often avoids synthetic issues with focus and selection.
Suggestion: Consider switching to @testing-library/user-event for interactions like typing, backspace and paste to make tests more robust.

**Suggestion:**
```typescript
Replace fireEvent.change with userEvent.type(inputs[0], '1') and use userEvent.keyboard('{Backspace}') etc.
```

src/component-catalog/InputOTP/InputOTP.tsx
```diff
@@ -0,0 +1,55 @@
+import { Box } from '@mui/material';
+
+import { DEFAULT_OTP_LENGTH, OTP_INPUT_PATTERN } from './InputOTP.constants';
+import { useInputOTP } from './InputOTP.hook';
+import { Container, StyledTextField, Label } from './InputOTP.styles';
+
+import type { InputOTPProps } from './InputOTP.types';
+
+export function InputOTP({
```
Ditmar, line 10: high: InputOTPProps (declared in InputOTP.types.ts) includes value and onChange indicating a controlled API, but the component ignores these props and uses internal state via useInputOTP. This is an API mismatch and will confuse consumers and tests.
Suggestion: Implement controlled behavior: if 'value' prop is provided, use it to populate inputs and call onChange when user edits. Alternatively, remove value/onChange from the props if controlled support is not desired.

**Suggestion:**
```typescript
Add logic to accept a controlled `value` (string) and `onChange` callback. For example, if value is provided, derive `values` from value and call onChange when digits change; otherwise use internal state from hook. Update useInputOTP to support controlled usage or move logic into the component.
```

src/component-catalog/InputOTP/InputOTP.tsx
```diff
@@ -0,0 +1,55 @@
...
+  length = DEFAULT_OTP_LENGTH,
+  disabled = false,
+  error = false,
+  onComplete,
+  label,
+}: InputOTPProps) {
+  const { values, inputRefs, handleChange, handleKeyDown, handlePaste } = useInputOTP(
```
Ditmar, line 17: medium: The component calls useInputOTP(length, disabled, onComplete) but does not memoize handlers passed to children. It passes inline arrow functions for onChange/onKeyDown/onPaste and inputRef assignment inside map which violates the team's rule against inline anonymous functions inside JSX where avoidable.
Suggestion: Either move per-index handlers into the hook and return bound callbacks or use useCallback to provide stable handlers that accept an index (e.g. handleChangeForIndex(index)). For inputRef assignment, consider using a callback returned from the hook.

**Suggestion:**
```typescript
Change hook to provide functions like getOnChange(index) returning a stable memoized handler, or create memoized wrapper callbacks in the component via useCallback to avoid new functions on each render.
```

src/component-catalog/InputOTP/InputOTP.tsx
```diff
@@ -0,0 +1,55 @@
...
+    length,
+    disabled,
+    onComplete
+  );
+
+  return (
+    <Container role="group" aria-label="One-time password input">
+      <Label>{label || 'Enter OTP Code'}</Label>
```
Ditmar: low: The default label string 'Enter OTP Code' is hardcoded in the component which may require localization. Project convention usually avoids hardcoded user-facing strings inside components.
Suggestion: Extract default labels and user-facing strings into a constants file or accept a label prop required and document default in docs, or use i18n/tokens.

**Suggestion:**
```typescript
Move the default text to a constant or to i18n, e.g. const DEFAULT_LABEL = 'Enter OTP Code' and use that constant in the component.
```

src/component-catalog/InputOTP/InputOTP.types.ts
```diff
@@ -0,0 +1,9 @@
+export interface InputOTPProps {
+  length?: number;
+  value?: string;
```
Ditmar: high: InputOTPProps declares `onChange?: (value: string) => void` and `value?: string` which implies a controlled component, but the implementation (InputOTP.tsx) does not use them. This is a contract mismatch and violates the team's requirement for explicit prop typing and clear component contracts.
Suggestion: Either implement controlled support in the component/hook or remove these props from the type definition. Document the component's controlled vs uncontrolled behavior in the .docs.mdx.

**Suggestion:**
```typescript
If controlled support is required, implement `value` handling and call `onChange` on each change. If not required, remove `value` and `onChange` from InputOTPProps.
```

Ditmar (changes requested):
---

**Conclusion:** There are multiple blocking violations of project conventions and technical issues: test-runner mixup (jest + vitest), missing React type imports, forbidden spread operator usage, MUI-internal class reliance, and mismatch between declared props and implementation. Address the high-severity issues, align testing stack, implement controlled API or update types, fix hook behavior for dynamic length, and correct Storybook MDX and style patterns before re-review.

package.json
```diff
@@ -48,9 +48,10 @@
     "@storybook/addon-vitest": "9.1.10",
     "@storybook/blocks": "9.0.0-alpha.17",
     "@storybook/react-vite": "9.1.10",
-    "@testing-library/jest-dom": "6.9.1",
-    "@testing-library/react": "16.3.0",
+    "@testing-library/jest-dom": "^6.9.1",
```
Ditmar: high: The change introduces caret (^) in versions for @testing-library/jest-dom and @testing-library/react while the rest of this file appears pinned. This creates inconsistency in dependency versioning strategy. Prefer consistent pinning or caret usage across devDependencies and document the policy in the repo.
Suggestion: Align the version ranges with the project's convention (either pin all devDependencies or use caret consistently).
Suggestion: Change to pinned versions consistent with the rest of package.json or update all devDependencies to caret format and add a comment in repository conventions stating the policy.

package.json
```diff
@@ -48,9 +48,10 @@
...
+    "@testing-library/react": "^16.3.0",
     "@testing-library/user-event": "14.6.1",
+    "@types/jest": "^30.0.0",
```
Ditmar: high: Adding @types/jest here suggests the project intends to use Jest typings; however the repository already has vitest listed (line 25). Mixing jest/@types/jest/ts-jest and vitest is confusing and may produce type and test runner conflicts.
Suggestion: Decide on a single test runner (Vitest or Jest). If using Vitest, remove @types/jest and ts-jest and rely on vitest's types (or add appropriate types). If moving to Jest, remove Vitest and migrate test config.
Suggestion: Remove @types/jest if continuing with Vitest. If switching to Jest, remove vitest and update CI/test scripts accordingly.

package.json
```diff
@@ -70,10 +71,12 @@
     "eslint-plugin-import": "2.32.0",
     "eslint-plugin-prettier": "5.5.4",
     "eslint-plugin-storybook": "9.1.10",
+    "jest": "^30.3.0",
```
Ditmar: 3. package.json:17
high: A new dependency 'jest' is added while vitest is already present (line 25). This will create duplication and confusion in CI and local test runs and is a breaking architectural decision for the testing setup.
Suggestion: Keep a single test framework. If choosing Jest, migrate tests and CI to Jest and remove vitest. If keeping Vitest, do not add jest as a devDependency.
Suggestion: Remove the added 'jest' entry and depend solely on vitest, or convert the rest of the codebase/tests and CI to jest and remove vitest.

package.json
```diff
@@ -70,10 +71,12 @@
...
     "jsdom": "27.0.0",
     "playwright": "1.55.1",
     "prettier": "3.6.2",
     "storybook": "9.1.10",
+    "ts-jest": "^29.4.9",
```
Ditmar: 4. package.json:22
high: ts-jest is added (line 22) which is only needed when using Jest with TypeScript. If the project uses Vitest (line 25), ts-jest is unnecessary. This couples testing stack incorrectly.
Suggestion: Remove ts-jest unless you intentionally move to Jest. Use Vitest-specific TypeScript setup instead.
Suggestion: Remove ts-jest and related jest packages if continuing with Vitest. If migrating to Jest, add a migration plan and update test scripts/config.

Ditmar (changes requested):

src/component-catalog/InputOTP/InputOTP.constants.ts
rodrigopaco: This regex intentionally allows empty values to support deletion/backspace behavior in the OTP inputs. Added a comment to clarify this.

src/component-catalog/InputOTP/InputOTP.docs.mdx
rodrigopaco: Fixed. Added proper Storybook MDX imports and connected the documentation to InputOTP stories using Meta, Canvas, and Story blocks.

src/component-catalog/InputOTP/InputOTP.docs.mdx
rodrigopaco: Fixed. Replaced invalid Storybook usage with valid MDX/Storybook blocks so the documentation renders correctly.

src/component-catalog/InputOTP/InputOTP.docs.mdx
rodrigopaco: Fixed. Updated the styles to remove hardcoded font sizes and replaced them with theme typography tokens.

This aligns the implementation with the documentation, which now correctly reflects the use of theme-based values.

src/component-catalog/InputOTP/InputOTP.hook.ts
rodrigopaco: Fixed. Imported explicit React event types using:

import type { ChangeEvent, KeyboardEvent, ClipboardEvent } from 'react';

and updated all handler signatures accordingly to avoid relying on global React types.

src/component-catalog/InputOTP/InputOTP.hook.ts
rodrigopaco: Fixed. Added a useEffect that resets both values and inputRefs when length changes:

useEffect(() => {
  setValues(Array(length).fill(''));
  inputRefs.current = new Array(length).fill(null);
}, [length]);

This ensures the hook stays in sync with dynamic length updates.

src/component-catalog/InputOTP/InputOTP.hook.ts
rodrigopaco: Fixed together with the length synchronization. inputRefs.current is now reset inside a useEffect whenever length changes, ensuring the ref array always matches the expected size.

src/component-catalog/InputOTP/InputOTP.hook.ts
rodrigopaco: Fixed. Replaced all spread usages ([...values]) with values.slice() to comply with project constraints.

src/component-catalog/InputOTP/InputOTP.hook.ts
rodrigopaco: Fixed. Updated the signature to use a properly typed event:

KeyboardEvent<HTMLInputElement>

after importing the type from React.

src/component-catalog/InputOTP/InputOTP.hook.ts
rodrigopaco: Addressed. Added an explanatory comment before preventDefault:

// Prevent default paste so we can sanitize and distribute pasted digits across inputs

to clarify the intended behavior.

src/component-catalog/InputOTP/InputOTP.stories.tsx
rodrigopaco: Adjusted based on ESLint import/order rules in the project, which require internal imports before external ones to pass lint checks.

src/component-catalog/InputOTP/InputOTP.stories.tsx
rodrigopaco: Fixed. Extracted the decorator into a named function (CenterDecorator) and replaced inline styles with a MUI Box-based layout.

This removes the inline anonymous function and avoids hardcoded style objects, aligning with project conventions. The layout is now handled using MUI props for better readability and maintainability.

src/component-catalog/InputOTP/InputOTP.styles.ts
rodrigopaco: Replaced hardcoded fontSize and lineHeight values with theme typography tokens (body2) to align with design system conventions and improve consistency.

src/component-catalog/InputOTP/InputOTP.styles.ts
rodrigopaco: Adjusted container layout to avoid enforcing centering and layout decisions. This allows parent components or story wrappers to control positioning, improving reusability.

src/component-catalog/InputOTP/InputOTP.styles.ts
rodrigopaco: Replaced decimal spacing value with theme.spacing(6) to maintain consistency with design tokens and improve readability.

src/component-catalog/InputOTP/InputOTP.styles.ts
rodrigopaco: Refactored styles to reduce reliance on MUI internal classes. 
Replaced usage of OutlinedInput-specific selectors with a more stable InputBase-level styling approach.

Where possible, styling has been simplified and aligned closer to MUI's public APIs to improve maintainability and reduce coupling.

src/component-catalog/InputOTP/InputOTP.styles.ts
rodrigopaco: Added a safe fallback for fontSize using optional chaining and body1 as fallback to ensure compatibility across different theme configurations.

src/component-catalog/InputOTP/InputOTP.test.tsx
rodrigopaco: Updated tests to align with Vitest by using vi.fn() for mocks and added coverage for onComplete when the OTP is completed by typing and paste.

src/component-catalog/InputOTP/InputOTP.test.tsx
rodrigopaco: Updated interaction tests to use userEvent for typing and keyboard behavior. fireEvent is still used for paste because the component handles clipboardData directly.

src/component-catalog/InputOTP/InputOTP.tsx
rodrigopaco: Resolved by aligning the component API with an uncontrolled approach.

The `value` and `onChange` props were removed from `InputOTPProps` since the component manages its own internal state via the `useInputOTP` hook. This ensures consistency between the public API and the implementation, avoiding confusion for consumers and tests.

The uncontrolled behavior is now documented and reflected across the component, hook, and types.

src/component-catalog/InputOTP/InputOTP.tsx
rodrigopaco: Resolved by memoizing handlers using useCallback.

Instead of inline arrow functions inside JSX, stable handler wrappers were created (e.g. handleChangeAt, handleKeyDownAt, handlePasteAt, setInputRef). This avoids recreating functions on each render and follows project conventions regarding performance and readability.

src/component-catalog/InputOTP/InputOTP.tsx
rodrigopaco: Resolved by extracting the default label into a constant.

The hardcoded string 'Enter OTP Code' was moved to a constant (DEFAULT_OTP_LABEL) inside InputOTP.constants.ts. The component now uses this constant as a fallback when no label prop is provided, improving maintainability and enabling easier localization in the future.

src/component-catalog/InputOTP/InputOTP.types.ts
rodrigopaco: Resolved by removing `value` and `onChange` from InputOTPProps and aligning the component with an uncontrolled API.

The component now manages its own internal state via the hook and exposes `onComplete` as the only callback when the OTP is fully entered.

The documentation will be updated to explicitly state that this is an uncontrolled component and describe its behavior accordingly.

package.json
rodrigopaco: Resolved. Removed caret ranges and aligned the dependency versions with the repository’s pinned version convention.

package.json
rodrigopaco: Resolved. Removed @types/jest because the project uses Vitest as the test runner, avoiding Jest/Vitest type conflicts.

package.json
rodrigopaco: Resolved. Removed ts-jest since it is only needed for Jest with TypeScript. The project continues using Vitest.

package.json
rodrigopaco: Resolved. Removed jest from devDependencies and kept Vitest as the single test framework used by the project.

Ditmar (approved):

PR #304, opened 2026-06-13
feat: add base Astro layout

rodrigopaco:
Implemented the base Astro layout used by all pages.

Includes:
- Default SEO metadata
- Per-page title and description support
- Canonical URL from Astro.url
- Open Graph tags
- Optional ogImage support
- noIndex support for private pages
- Named head slot for extra page metadata
- Font loading through HTML links
- Global CSS reset and baseline styles
- Updated existing pages to use page-specific metadata

Validation:
- yarn lint passes
- yarn build passes
- yarn test passes


src/layouts/BaseLayout.astro
```diff
@@ -4,116 +4,105 @@ import '../style-library/theme/colors.css';
 interface Props {
   title?: string;
   description?: string;
+  ogImage?: string;
+  noIndex?: boolean;
```
Ditmar, line 8: The default description has been changed to a more specific one, which is good for SEO. However, ensure that this aligns with the overall branding and messaging strategy.

**Suggestion:**
```typescript
Confirm that the new default description is approved and aligns with the project's goals.
```

src/layouts/BaseLayout.astro
```diff
@@ -4,116 +4,105 @@ import '../style-library/theme/colors.css';
...
 }
 
 const {
-  title = 'LeetCode SPA',
-  description = 'Interactive coding practice platform',
+  title = 'LeetCode Clone',
```
Ditmar, line 13: The addition of the canonical URL is a good practice for SEO, but ensure that the URL generated by Astro.url is correct and reflects the current page context.

**Suggestion:**
```typescript
Test the canonical URL generation to ensure it works as expected.
```

src/layouts/BaseLayout.astro
```diff
@@ -4,116 +4,105 @@ import '../style-library/theme/colors.css';
...
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta name="theme-color" content={themeColor} />
     <meta name="description" content={description} />
-    <meta property="og:title" content={title} />
-    <meta property="og:description" content={description} />
     <meta name="generator" content={Astro.generator} />
-    <!-- MUI Emotion insertion point – must come before any other styles -->
+
```
Ditmar, line 31: The Open Graph tags are now included, which is great for social sharing. However, ensure that the ogImage prop is being passed correctly from all pages.

**Suggestion:**
```typescript
Verify that all pages using BaseLayout provide the ogImage prop when necessary.
```

src/layouts/BaseLayout.astro
rodrigopaco: Verified. The canonical URL is generated from Astro.url.href and reused for both the canonical link and og:url, so it reflects the current page context for each route.

src/layouts/BaseLayout.astro
rodrigopaco: Updated. I restored the default description to the original project messaging, "Interactive coding practice platform", to avoid unintended branding changes while still keeping the required SEO fallback description.

src/layouts/BaseLayout.astro
rodrigopaco: Verified. The ogImage prop remains optional and is rendered only when a page provides it. I did not add a default image because there is no confirmed shared Open Graph image for the project yet, but the layout is ready for any page that needs a specific social sharing image.

src/layouts/BaseLayout.astro
```diff
@@ -4,91 +4,127 @@ import '../style-library/theme/colors.css';
...
```
Ditmar, line 31: Ensure that the noIndex prop is applied consistently across all relevant pages.

**Suggestion:**
```typescript
Review all pages to confirm that the noIndex prop is applied where necessary.
```

rodrigopaco: Updated. I reviewed the current pages and found that /sysinfo was defining its own head outside BaseLayout. I updated it to use BaseLayout and applied the noIndex prop because it is an internal system information route. The remaining current routes are public pages.

src/layouts/BaseLayout.astro
```diff
@@ -4,91 +4,128 @@ import '../style-library/theme/colors.css';
...
+    <link rel="canonical" href={canonicalUrl} />
     <link rel="icon" type="image/svg+xml" href="/favicon.svg" />
+
+    <!-- Application font is loaded through HTML links, not client JavaScript. -->
+    <link rel="preconnect" href="https://fonts.googleapis.com" />
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
+    <link
+      rel="stylesheet"
+      href="https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600;700;800&display=swap"
```
Ditmar, line 52: Esto trae 'Inter' desde Google Fonts (CDN externo), mientras que todas las demas fuentes del proyecto (Poppins, Montserrat, Syne) estan self-hosted en /fonts como @font-face mas abajo en este mismo archivo. Ademas de la inconsistencia, suma una dependencia de red externa (con su costo de performance y el tema de exposicion de IP a Google). Si el objetivo es soportar Inter, lo mas prolijo seria bajar los woff2 y agregarlos al mismo bloque @font-face de aca abajo.

rodrigopaco: Thank you for the suggestion.

Inter is currently loaded through the external `<link>` because the LAY-001 acceptance criteria explicitly require the application font to use `<link rel="preconnect">` and `<link rel="stylesheet">`, rather than being bundled through JavaScript.

The existing self-hosted Poppins, Montserrat, and Syne fonts remain available in the layout. Since self-hosting Inter would require adding new font assets and expanding the scope of this already approved ticket, I kept the current implementation. This can be addressed separately as a font-loading and privacy improvement.

